Issue #3, opened 2021-07-14
Add attribution to map


johnrees:
can be hardcoded for now

© Crown copyright and database rights 2021 OS (0)100019252.

jessicamcinchak:
closed via #4

PR #4, opened 2021-07-14
feat: clickable feature layer, draw mode control, and attribution

### LINKED ISSUE #2: Clickable features
Example here - https://labs.os.uk/public/os-data-hub-examples/os-features-api/wfs-example-get-feature

Their code is a bit nasty but we can do a first pass with it. Maybe just disable the drawing tool for now while doing initial experiments.

I'll get the API key over to you @jessicamcinchak, it's a separate key to the OS datahub key that we're using for the map base layer

jessicamcinchak:
closes #2 & #3

known bugs:
- you can still click underlying features while drawing! it's weird for sure. a few possible ways come to mind to address: 
  - when you enter draw mode, we could disable the single-click-to-select-feature event listener so each point of your polygon doesn't select it's underlying feature. then we'd presumably want a way to re-enable this when you finish drawing (which we don't really have a way of knowing yet)
  - create a custom control for selecting features the same way we have for drawing & only allow a single mode at a time. in this case, if you click a feature and then switch to drawing, would we want to clear the map or keep the selection so you can "trace" ? and vice-versa, what should happen between interactions? 
- no popup associated with the features when you click them - [OS examples](https://labs.os.uk/public/os-data-hub-examples/os-features-api/wfs-example-get-feature) are still on open layers API v4, while current version is v6 - so Popup has evolved into [Overlay](https://openlayers.org/en/latest/apidoc/module-ol_Overlay-Overlay.html) and we'll need to look into how to adapt [this example](https://openlayers.org/en/latest/examples/overlay.html), I think


johnrees (approved):
don't worry about the known bugs for now, we're probably not going to use popups and the drawing/selecting thing can be completely separate.

We can ignore drawing atm and come back to it.. it's helpful having the button, thanks!

PR #8, opened 2021-07-21
feat: calculate area of polygons

jessicamcinchak:
- [x] for custom drawings
- [x] for "modified" drawings
- [x] for features clicked
- [x] for unioned features clicked

just console logging the area measurements for now until we make more design decisions! open layers ["measure" examples](https://openlayers.org/en/latest/examples/measure.html) use dynamic tooltips directly on the map, while planx currently renders total area outside of map container.


johnrees:
LGTM! Thanks! 📐

PR #13, opened 2021-07-29
dispatch events

johnrees:
maybe events should be props, idk

we should probably also keep area in the class like you had before, i'm just trying to keep the ui as simple as possible atm

### DIFF
--- a/index.html
+++ b/index.html
@@ -22,8 +22,8 @@
       map.addEventListener("ready", (event) => {
         console.log("map ready");
       });
-      map.addEventListener("areaChange", (event) => {
-        console.log({ area: event.detail });
+      map.addEventListener("areaChange", ({ detail: area }) => {
+        console.debug({ area });
       });
     </script>
   </body>
--- a/index.html
+++ b/index.html
@@ -25,6 +25,9 @@
       map.addEventListener("areaChange", ({ detail: area }) => {
         console.debug({ area });
       });
+      map.addEventListener("geojsonChange", ({ detail: geojson }) => {
+        console.debug({ geojson });
+      });
     </script>
   </body>
 </html>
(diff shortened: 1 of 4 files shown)


johnrees:
(it console.logs area, but will spam you when you resize a polygon)

index.html
```diff
@@ -15,5 +15,19 @@
...
+
+      map.addEventListener("ready", (event) => {
+        console.log("map ready");
+      });
+      map.addEventListener("areaChange", ({ detail: area }) => {
+        console.debug({ area });
+      });
+      map.addEventListener("geojsonChange", ({ detail: geojson }) => {
+        console.debug({ geojson });
+      });
```
jessicamcinchak, line 30: oh, cool! :+1:

src/my-map.ts
```diff
@@ -107,30 +113,46 @@ export class MyMap extends LitElement {
...
 
   // render the map
   render() {
     return html`<script src="https://cdn.polyfill.io/v2/polyfill.min.js"></script>
       <link rel="stylesheet" href="https://cdn.skypack.dev/ol@^6.6.1/ol.css" />
-      <div id="map" />
-      ${this.drawMode
-        ? html`<span id="area">${this.totalArea}</span>`
-        : null} `;
+      <div id="map" />`;
```
jessicamcinchak, line 142: totally agree map should really only be returning the map -- while doing this I was thinking it'd be interesting to return the totalArea that we calculate _in_ map as it's own `@customElement("total-area")` span that's easy to embed separately later or just omit, but have no idea if that's possible or how'd it work yet!

johnrees: oh yeah that's an interesting idea, would be fun to explore

PR #15, opened 2021-07-30
feat: reset control to clear drawing

### LINKED ISSUE #6: Disable map rotation on touch devices

jessicamcinchak:
custom control button to reset to original zoom & center and clear any drawn features if they exist.

also disabling rotation which should fix #6. 

maybe this should be dispatching an event like #13 though?

### DIFF
--- a/index.html
+++ b/index.html
@@ -4,7 +4,7 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Map</title>
-    <script type="module" src="src/my-map.ts"></script>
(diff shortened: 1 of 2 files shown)


johnrees:
good thinking! merge anytime, sorry for creating the conflict! Don't worry about events for that for now. The main one will be when a polygon has been drawn I think, so we can click continue in the DrawBoundary component

johnrees:
good thinking! merge anytime, sorry for creating the conflict! Don't worry about events for that for now. The main one will be when a polygon has been drawn I think, so we can click continue in the DrawBoundary component

PR #16, opened 2021-07-30
fix: limit drawing to a single polygon

jessicamcinchak:


johnrees:
nice, thanks ✏️

jessicamcinchak:
got a little complicated by reset control, so keeping this one open for a bit longer!

PR #24, opened 2021-08-06
feat: query & display features that intersect with lon,lat

jessicamcinchak:
creeping into the "future" features with this one, but the logic was basically already on hand from our first implementation! this will let us show a border around the address point in FindProperty component & move towards having a more accurate polygon/geojson to make more accurate planning constraint queries in the future.

example implementation:
```html
<my-map latitude="51.4858363" longitude="-0.0761246" showFeaturesAtPoint featureColor="#8a2be2" />
```

### DIFF
--- a/src/my-map.ts
+++ b/src/my-map.ts
@@ -66,6 +66,9 @@ export class MyMap extends LitElement {
   @property({ type: String })
   featureColor = "#0000ff";
 
+  @property({ type: Number })
+  featureBuffer = 40;
+
   private useVectorTiles =
     Boolean(import.meta.env.VITE_APP_ORDNANCE_SURVEY_KEY) &&
     osVectorTileBaseMap;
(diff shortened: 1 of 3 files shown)


src/my-map.ts
```diff
@@ -56,7 +58,16 @@ export class MyMap extends LitElement {
   maxZoom = 22;
 
   @property({ type: Boolean })
-  drawMode = true;
+  drawMode = false;
```
jessicamcinchak, line 61: this is a slightly extraneous, but key change! made the decision here that our various feature modes should each be "off" by default. once this is merged & published, we should update the codesandbox demo to explicitly turn on 'drawMode'.

johnrees: Yeah that makes sense!

If this gets published we'll need to notify bops of this change too, or get them to work with a previous fixed version number for now until the API is a bit more settled

johnrees (approved):
Very coooooooool 😎

PR #25, opened 2021-08-10
wip: add storybook

johnrees:
I'm not really happy with this as it's adding a lot of dependencies and fluff. Might look at some simpler alternatives for the time being, even if it's static HTML files in an examples dir.


johnrees:
I'm think I'm going to close this PR @jessicamcinchak but the code for storybook is here if we ever need to reference or work from it

PR #27, opened 2021-08-10
upgrade from lit-element > lit

johnrees:
I think lit-element is deprecated now and lit is a monorepo that we should be tracking?


jessicamcinchak (approved):
Yep, lit seems like the way to go now!

PR #29, opened 2021-08-10
feat: provide OS API keys as properties, rather than env vars

### LINKED ISSUE #21: Provide OS keys via props rather than ENV vars

jessicamcinchak:
Adds configurable properties for specifying OS API keys.

Also cleans up naming throughout to hopefully make OS-related services a bit more clear & consistent. I've defined a property per OS API for now, even though some end users may have a single key that can access both/any.

Closes #21


src/my-map.ts
```diff
@@ -84,17 +85,27 @@ export class MyMap extends LitElement {
   @property({ type: Number })
   geojsonBuffer = 12;
 
-  private useVectorTiles =
-    Boolean(import.meta.env.VITE_APP_ORDNANCE_SURVEY_KEY) &&
-    osVectorTileBaseMap;
```
jessicamcinchak, line 89: this was a bit confusing: I originally thought `useVectorTiles` could use lit's `@state()` decorate as an internal reactive property, but since it now references other configurable property values, it has to be defined within `firstUpdated()` to evaluate correctly!

jessicamcinchak:
Netlify failing because env vars have been renamed & are not yet updated under "deploy settings" -- will fix after review depending if we decide to keep env vars as default prop values or remove them altogether!

jessicamcinchak:
Update: now using `env.development.local` instead of `.env.local` - as this prevents our personal API keys from getting bundled with the production build (when running `ENVIRONMENT=production pnpm build` at least). 

A perk of not removing env vars all together is that we can still set them on Netlify deploy settings this way, so that our OS preview links have all features & styles available. 

End users should still specify all keys as props. I cleaned up the readme to remove mention of env vars in examples!

johnrees (approved):
Woo! Great stuff! A+ would review again

PR #30, opened 2021-08-10
feat: npm publish main if package version changes

johnrees:
Currently set to dry-run for testing. I'm not sure if it'll like pnpm yet. GitHub is being weird too so I'll try this out tomorrow.

In theory we can keep merging to main as we are, then when we want to create a release, we create a PR that bumps the package version number (and any CHANGELOG updates) and it'll then auto publish on merge.

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,19 +1,35 @@
 name: Publish to NPM
 on:
-  push:
+  # to REMOVE before merge to main
+
+  pull_request:
+    types: [opened, synchronize]
     branches:
       - main
+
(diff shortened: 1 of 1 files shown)


johnrees:
netlify failed because github was dead at the time

jessicamcinchak:
cool!! this seems like a great simple solution

i queued up some changes to try it out!! :https://github.com/theopensystemslab/map/pull/36 (but let's merge #34 first :slightly_smiling_face:)

johnrees:
closing in favour of #135 but happy to pursue this route instead if you think it's better @jessicamcinchak

PR #50, opened 2021-08-20
fix: ensure prior drawingSource features are cleared when entering drawMode

### LINKED ISSUE #46: Empty drawing layer features before each draw
There's a bit of weird "back" behavior in DrawBoundary where if you haven't "refreshed", but just gone "back", the DrawBoundary map may initially load the last drawn boundary, and then still let you draw exactly one more. `setBoundary` will always look to store the first feature, which in this case would be the one that pre-loaded, not the fresh drawing. Clicking "refresh" works as expected and clears everything.

Making sure the layer is emptied before you begin drawing should clear this up.

jessicamcinchak:
Fixes #46

### DIFF
--- a/src/my-map.ts
+++ b/src/my-map.ts
@@ -228,6 +228,7 @@ export class MyMap extends LitElement {
     }
 
     if (this.drawMode) {
+      drawingSource.clear();
       map.addLayer(drawingLayer);
 
       map.addInteraction(modify);


src/my-map.ts
```diff
@@ -228,6 +228,7 @@ export class MyMap extends LitElement {
     }
 
     if (this.drawMode) {
+      drawingSource.clear();
```
gunar, line 231: Is imperative the only way to solve this? Just checking. And if then, do you think it might be interesting to add a comment here explaining why `.clear()` is even necessary at this point?

gunar (approved):
Love how organized you are with GitHub issues and PRs ✨

PR #51, opened 2021-08-20
feat: `areaUnits` as configurable property

### LINKED ISSUE #49: Make total area units a configurable property
Options should match params of this utility function: https://github.com/theopensystemslab/map/blob/main/src/utils.ts#L7

Wonder if there's a smart way to do ENUM type properties in Lit?

jessicamcinchak:
Closes #49

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,13 +1,15 @@
 import Geometry from "ol/geom/Geometry";
 import { getArea } from "ol/sphere";
 
+export type AreaUnitsEnum =  "m2" | "ha";
+
 /**
  * Calculate & format the area of a polygon
  * @param polygon
  * @param units - defaults to square metres ("m2"), or supports "ha" for hectares
- * @returns - a string
+ * @returns - the total area formatted with units as a string
  */
-export function formatArea(polygon: Geometry, units: String = "m2") {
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -1,13 +1,15 @@
 import Geometry from "ol/geom/Geometry";
 import { getArea } from "ol/sphere";
 
+export type AreaUnitsEnum =  "m2" | "ha";
```
gunar, line 4: I think you might be thinking of "units" in terms of "5 units of m2" but I'd like to suggest a different perspective—that of a "unit of magnitude." From this point of view, the name "unit" should be singular in both `AreaUnitsEnum` and `areaUnits` as the consumer can pick one and no more than one unit of measurement.

That is:
- `AreaUnitsEnum` => `AreaUnitEnum`
- `areaUnits` => `areaUnit`

The reason I bring this up is because this is now a published interface so changing the name later on would mean releasing a breaking change and a new major version. Even though no one is really using it, it might be interested to start getting into that mindset.

What do you think?

gunar (commented):
One comment. Everything else looks good.

gunar (approved):

PR #57, opened 2021-09-12
feat: show vertices of drawn polygon

jessicamcinchak:
- Show vertices of drawn polygon once a shape has been closed; user can modify existing vertices or add new ones (design similar to MapInfo Professional used internally at Buckinghamshire)
- Tweak default drawing style to have more prominent fill color
- Increase snap pixel tolerance against drawing layer so it's a bit easier to close your drawing & grab an existing vertex to modify

Screenshot from 2021-09-12 13-54-08

### DIFF
--- a/index.html
+++ b/index.html
@@ -14,7 +14,7 @@
     />
   </head>
   <body>
-    <my-map zoom="18" drawMode />
+    <my-map zoom="18" maxZoom="20" drawMode />
 
     <script>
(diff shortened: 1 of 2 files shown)


jessicamcinchak:
[Snap to OS base map layer?](https://trello.com/c/KCJSls1W/1502-snap-to-os-base-map-layer)

melzreal (approved):
LGTM :shipit:

Issue #37, opened 2021-08-13
Boolean properties should be `false` by default


jessicamcinchak:
Lit seems to treat boolean type properties by presence, rather than falsy-ness of value - which makes it confusing to try and set a property that is true by default to `renderVectorTiles=“false”`.

For now, these can all just be false by default.

Issue #26, opened 2021-08-10
Add different examples


johnrees:
Related https://github.com/theopensystemslab/map/pull/25

jessicamcinchak:
a few recent ideas around this:

non-storybook:
- https://component.kitchen/blog/posts/building-a-storybook-like-demo-browser-with-web-components-a-much-simpler-way-to-get-most-of-the-benefits
- https://pitsby.com/
- https://dev.to/webpadawan/api-viewer-document-your-web-components-api-p8j

storybook:
- build with vite? https://storybook.js.org/blog/storybook-for-vite/

PR #60, opened 2021-09-21
Add scale bar to map

melzreal:
### Description of change

Add scale bar to maps
image

### Story Link

https://trello.com/c/3LkxI6Kd/448-map-scale-bar-required-on-decision-notice

Co-authored-by: Rhian Lewis <rhianatwork@gmail.com>

### DIFF
--- a/src/my-map.ts
+++ b/src/my-map.ts
@@ -12,6 +12,7 @@ import View from "ol/View";
 import { last } from "rambda";
 
 import { draw, drawingLayer, drawingSource, modify, snap } from "./draw";
+import { scaleControl } from "./scale-line"
 import {
   makeFeatureLayer,
   outlineSource,
@@ -115,6 +116,12 @@ export class MyMap extends LitElement {
 
   @property({ type: String })
   ariaLabel = "Interactive map";
+  
+  @property({ type: Boolean })
+  showScale = false;
+
+  @property({ type: Boolean })
+  useScaleBarStyle = false;
 
   // runs after the initial render
   firstUpdated() {
@@ -148,7 +155,7 @@ export class MyMap extends LitElement {
       controls: defaultControls({
         attribution: true,
         zoom: !this.staticMode,
-      }),
+      }).extend(this.showScale ? [scaleControl(this.useScaleBarStyle)] : []),
       interactions: defaultInteractions({
         doubleClickZoom: !this.staticMode,
         dragPan: !this.staticMode,
--- a/src/scale-line.ts
+++ b/src/scale-line.ts
@@ -1,0 +1,13 @@
+import 'ol/ol.css';
+import { ScaleLine } from 'ol/control';
+
+export function scaleControl(useScaleBarStyle: boolean) {
+  return new ScaleLine({
+    units: 'metric',
+    bar: useScaleBarStyle,
+    steps: 4,
+    text: true,
+    minWidth: 140,
+  });
+}
+


src/my-map.ts
```diff
@@ -142,7 +143,7 @@ export class MyMap extends LitElement {
       controls: defaultControls({
         attribution: true,
         zoom: !this.staticMode,
-      }),
+      }).extend([scaleControl()]),
```
jessicamcinchak, line 146: this scale bar is a super handy visual feature for map orientation! my only suggestion is that we don't make it visible by default all the time, but rather add a [Lit property](https://lit.dev/docs/components/properties/) so that end users can opt in or out of showing it for various use cases/implementations.

one way to do that could be like this: 
1. in `my-map.ts` around L114, add:
```ts
@property({ type: Boolean })
showScale = false;
```
configurable properties are just a long unordered list with most recent additions at the bottom currently, open to organization ideas here! current rule of thumb is that boolean properties are always false by default, see [here](https://github.com/theopensystemslab/map/issues/37) for context

2. then this line could use a conditional ternary operator to _only_ add the scale when the property is true:  
```ts
...}).extend(this.showScale ? [scaleControl()] : []),
```

3. finally, in our individual implementations of this component, we would specify that we want to show the scale bar in our html element like this:
```html
<my-map zoom="18" drawMode showScale />
```
omitting the `showScale` property would simply hide it

src/scale-line.ts
```diff
@@ -0,0 +1,16 @@
+import 'ol/ol.css';
+import { ScaleLine } from 'ol/control';
+
+let control;
+
+export function scaleControl() {
+  control = new ScaleLine({
```
jessicamcinchak, line 7: this could just be `return new ScaleLine({` and & then you wouldn't have to declare `control` globally above on line 4 :)

jessicamcinchak (changes requested):
very exciting & useful first feature contribution :tada:

left a couple comments, happy to talk through if anything is unclear!

src/my-map.ts
melzreal: thanks this is super thorough! Sounds good to me - I've put the question to the partners since the ticket we had seems to suggest they always need a scale bar for legal purposes, so it seemed logical to just enforce it everywhere - but actually it is probably best that the base map just has that option as you suggest and then we turn it on everywhere we need it - just double checking what they want

jessicamcinchak: awesome, glad this worked out! on our side, toggling it on/off will be super helpful as we're often trying to keep the public-facing applicant maps as least cluttered & technical as possible. 

will approve this now & once you merge it I'll add it to the next release I have queued up [here](https://github.com/theopensystemslab/map/pull/59) -- which we can plan to merge/publish first thing tmrw morning :slightly_smiling_face: :rocket:

jessicamcinchak (approved):

src/my-map.ts
melzreal: thanks for the approval! It occurred to me today though that instead of hard coding the bar type they decided on, we could instead pass it on as an argument since Os does provide two different bar types - just in case they change their mind and then want the other one..

melzreal: (will figure out how to do it and incorporate) - also should I update documentation somewhere? Was wondering how would anyone apart from us know that showScale can be turned on and off

jessicamcinchak: cool! the bar style seems like a nice argument to have available for more custom implementation options. of course feel to keep working even though i approved, just re-request when you're done i guess?

as for documentation, i've been adding examples direct to the README for now! there's also [this issue](https://github.com/theopensystemslab/map/issues/26) to figure out a longer term solution, maybe Storybook or something similar. we're still pre-v1, so can stay lightweight & flexible about this for now in my opinion until we settle on a solution we like :slightly_smiling_face:

src/scale-line.ts
```diff
@@ -0,0 +1,20 @@
...
+      steps: 4,
+      text: true,
+      minWidth: 140,
+    });
+  } else {
+    return new ScaleLine({
+      units: 'metric',
+      bar: false,
+    });
+  }
```
jessicamcinchak, line 18: i think this if/else can be simplified to: 
```ts
export function scaleControl(showBar: boolean) {
  return new ScaleLine({
    units: 'metric',
    bar: showBar,
    steps: 4,
    text: true,
    minWidth: 140,
  });
}
```

small detail: typescript is lowercase boolean, while Lit property type is uppercase B ! Linter will complain about that.

src/my-map.ts
```diff
@@ -115,6 +116,12 @@ export class MyMap extends LitElement {
 
   @property({ type: String })
   ariaLabel = "Interactive map";
+  
+  @property({ type: Boolean })
+  showScale = false;
+
+  @property({ type: Boolean })
+  showBar = false;
```
jessicamcinchak, line 124: :+1: another property makes sense to me here! I do find this naming slightly confusing though, wonder if `useBarStyleScale` is more descriptive for the second one here? up to you though!

PR #73, opened 2021-11-01
fix: typo

gunar:


jessicamcinchak (approved):
thanks for catching :see_no_evil: !! will include this in the next release, but I'm going to wait to merge until then as it's nice to keep main level with current release in the meantime

PR #74, opened 2021-11-05
tidy up snaps

jessicamcinchak:
looks like more than it is because i'm based off main!

makes sure we only add the snapping layer when in drawMode with the vectorTile basemap enabled

shuffles the code around a bit to keep it closer to drawing logic in the flow of my-map.ts, and moves a couple more definitions to snapping.ts

not thinking about styling yet! maybe just easiest to set a snapping color property so it's easy to keep changing minds about? :upside_down_face: 

also still thinking about how snaps should work once you close your shape? can we actually modify-to-snaps? if we cant, maybe points should disappear?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 The minor version will be incremented upon a breaking change and the patch version will be
 incremented for features.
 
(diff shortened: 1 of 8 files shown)


jessicamcinchak:
going to merge into your PR so we have a single place for BOPs devs to code review & can unblock this ! https://trello.com/c/rQMerGkO/447-red-line-tool-snap-to-feature

PR #83, opened 2021-11-19
fix: keep snapping behaviour while modifying polygon

johnrees:
I think this should fix this issue with modify shape not snapping, apologies for the extra prettier noise! We'll get prettier running on everything in this repo again at some point.

Steps taken

- ensure snap interaction behaviour added _after_ draw and modify behaviours
- don't remove snap behaviour after completing shape

### DIFF
--- a/src/my-map.ts
+++ b/src/my-map.ts
@@ -10,7 +10,6 @@ import { Vector as VectorSource } from "ol/source";
 import { Fill, Stroke, Style } from "ol/style";
 import View from "ol/View";
 import { last } from "rambda";
-
 import { draw, drawingLayer, drawingSource, modify, snap } from "./drawing";
 import {
   getFeaturesAtPoint,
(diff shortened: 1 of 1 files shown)


src/my-map.ts
```diff
@@ -295,9 +294,11 @@ export class MyMap extends LitElement {
       if (!loadInitialDrawing) {
         map.addInteraction(draw);
       }
-      map.addInteraction(snap);
       map.addInteraction(modify);
 
+      // XXX: snap must be added after draw and modify
+      map.addInteraction(snap);
```
jessicamcinchak, line 300: oo this is interesting!! i had tried not removing the snap interaction, but hadn't thought to re-order how they're added! and figured because the modify interaction can only have a single source which is drawingSource, rather than the pointsSource, that it wasn't quite possible

jessicamcinchak (approved):

PR #86, opened 2021-11-24
break: remove ariaLabel property

jessicamcinchak:
Accessibility report says that aria-label attributes are not allowed on divs and may cause random results for assistive technology. It recommendeds removing them entirely.


jessicamcinchak:
[ARIA Attributes](https://trello.com/c/JJvHTvq5/1659-aria-attributes)

johnrees (approved):
thanks! [insert ariana grande/aria-not-WCAG-AA pun]

PR #93, opened 2021-12-02
feat: add `<postcode-search />` & `<address-autocomplete />` components

jessicamcinchak:
Transitioning the map into a small library of web components to work with UK address data via the Ordnance Survey APIs. 

Autocomplete is the main priority here so we can pass our accessibility audit :crossed_fingers:, but we don't actually have to be in any hurry to integrate the `<postcode-search />`, especially if overriding/merging styles proves tedious during integration. More context & motivation in [this discussion thread](https://github.com/theopensystemslab/map/discussions/92).

**Components**
1. `<postcode-search />` is a GOV.UK-styled input that validates UK postcodes using these [postcode utility methods](https://www.npmjs.com/package/postcode). I added [govuk-frontend](https://frontend.design-system.service.gov.uk/installing-with-npm/) & sass for styling; [this example repo](https://github.com/e111077/vite-lit-element-ts-sass) was handy for configuring sass with lit & vite. When a postcode is validated, an event is dispatched containing the sanitized string.
Screenshot from 2021-12-02 22-18-43Screenshot from 2021-12-02 22-18-29

2. `<address-autocomplete />` fetches addresses in a given UK postcode using the [OS Places API](https://apidocs.os.uk/docs/os-places-overview) and displays them using GOV.UK's [accessible-autocomplete](https://github.com/alphagov/accessible-autocomplete) component. When you select an address, an event is dispatched with the full OS record.
Screenshot from 2022-03-18 14-42-42

- Components are completely unaware of each other; we can dynamically render them together like our current FindProperty design pattern or they could be rendered on separate pages like a more traditional [GOV.UK address design pattern](https://design-system.service.gov.uk/patterns/addresses/)
- We should be able to override styles & further customize to fit in with our Material UI themes during integration without needing to directly introduce a sass dependency to planx

**Other changes**
- Refactored file structure to better reflect a library of components
- Reconfigured build (notes here #116)

**Future scope // Nice to haves**
- Global sass file or per component? Would be nice to take advantage of per-component GOV.uk style imports if possible. There are also so many sass deprecation warnings on build, and only importing components we need may help lighten that.
- We should finally think of a better name for all of this
- Better docs solution than current `index.html`, pick back up #61 ?
- Tests would be good, overdue to read up on how best to test web components


src/components/address-autocomplete/index.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    return !this.osPlacesApiKey || this._osError || this._totalAddresses === 0
+      ? html`<script src="https://cdn.polyfill.io/v2/polyfill.min.js"></script>
+          <div class="govuk-warning-text">
+            <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+            <strong class="govuk-warning-text__text">
+              <span class="govuk-warning-text__assistive">Warning</span>
+              ${errorMessage}
+            </strong>
+          </div>`
```
DafyddLlyr, line 159: This is nice!

index.html
```diff
@@ -14,9 +14,26 @@
...
+      <div stye="margin-bottom:1em">
+        <postcode-search hintText="Optional hint text shows up here" />
+      </div>
+      <div style="margin-bottom:1em;">
+        <!-- 
+          Examples (as of Feb 2022):
+          SE19 1NT (Lambeth): 56 DPA addresses -> 128 LPI addresses (87 "approved"), now requires paginated fetch
+          HP11 1BR (Bucks): 0 addresses, shows "No addresses found in postcode" error message
+          HP11 1BC (Bucks): valid postcode according to npm package but not OS, shows OS error message
+        -->
```
DafyddLlyr, line 31: 👍

index.html
```diff
@@ -14,9 +14,26 @@
...
   </head>
   <body>
-    <my-map zoom="20" maxZoom="23" drawMode drawPointer="dot" id="example-map" />
-
+    <div style="display:flex;flex-direction:column;">
+      <h1 style="color:red;font-family:Inter,Helvetica,sans-serif;font-size:16px;">*** This is a testing sandbox - these components are unaware of each other! ***</h1>
+      <div style="margin-bottom:1em">
+        <my-map zoom="20" maxZoom="23" drawMode drawPointer="dot" id="example-map" />
+      </div>
+      <div stye="margin-bottom:1em">
```
DafyddLlyr, line 22:
```suggestion
      <div style="margin-bottom:1em">
```

DafyddLlyr (changes requested):
Looking great! A few very minor comments on the postcode component, but no show stoppers 👍 

The unique id issue is something that shows up in a few places - we can ignore this for now if this is more pragmatic, but just put a card on the backlog to refactor when we hit pages?

src/components/postcode-search/index.ts
```diff
@@ -0,0 +1,134 @@
...
+        </span>
+        <input
+          class="govuk-input govuk-input--width-10"
+          id="postcode"
+          name="postcode"
+          type="text"
+          autocomplete="postal-code"
+          spellcheck="false"
+          aria-describedby="event-name-hint event-name-error"
+          .value=${this._postcode}
```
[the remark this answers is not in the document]
jessicamcinchak: ah so this dot is actually an intentional, significant Lit detail! Lit uses `.value` to assign the value property, whereas `value` would assign the attribute. I'll add a comment saying so, it definitely looks a bit unusual.

I found these links helpful for wrapping my head around some of these syntax quirks:
- https://levelup.gitconnected.com/build-a-material-like-input-web-component-using-litelement-20e9e0d203b6
- https://stackoverflow.com/questions/61430631/onchange-not-fired-in-input-field-of-litelement

src/components/postcode-search/index.ts
```diff
@@ -0,0 +1,134 @@
...
+      <div class="govuk-form-group">
+        ${this._makeLabel()}
+        <div id="event-name-hint" class="govuk-hint">${this.hintText}</div>
+        <span
+          id="event-name-error"
+          class="govuk-error-message"
+          style="display:none"
+        >
+          <span class="govuk-visually-hidden">Error:</span>${this.errorMessage}
+        </span>
```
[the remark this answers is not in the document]
jessicamcinchak: good catch - updated! additionally added `role="status"` to the autocomplete container to be safe, as we'll still plan to conditionally render that component.

src/components/postcode-search/index.ts
```diff
@@ -0,0 +1,134 @@
...
+        >`;
+  }
+
+  render() {
+    return html`<script src="https://cdn.polyfill.io/v2/polyfill.min.js"></script>
+      <div class="govuk-form-group">
+        ${this._makeLabel()}
+        <div id="event-name-hint" class="govuk-hint">${this.hintText}</div>
+        <span
+          id="event-name-error"
```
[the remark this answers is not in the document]
jessicamcinchak: good point - `id` & `errorId` are now user-configurable properties - so we can handle these custom cases & pages in the future!

src/components/postcode-search/index.ts
```diff
@@ -0,0 +1,134 @@
...
+      this._postcode = toNormalised(input.trim()) || input;
+    } else {
+      this._sanitizedPostcode = null;
+      this._postcode = input.toUpperCase();
+    }
+
+    if (this._sanitizedPostcode) {
+      this.dispatch("postcodeValidated", {
+        postcode: this._sanitizedPostcode,
+      });
```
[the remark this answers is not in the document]
jessicamcinchak: updated! much nicer visually :+1:

vite.config.ts
```diff
@@ -1,12 +1,15 @@
...
+import path from "path";
+import litcss from 'rollup-plugin-postcss-lit';
 
 // https://vitejs.dev/config/
 export default defineConfig({
   build: {
     lib: {
-      entry: "src/my-map.ts",
+      entry: path.resolve(__dirname, "src/index.ts"),
+      name: "ComponentLib",
```
jessicamcinchak, line 10: to doublecheck: I won't break any existing imports or the cdn script (eg https://cdn.jsdelivr.net/npm/@opensystemslab/map@0.4.0) by changing `build.lib.name`, right? imports are only reliant on the component names in `index.ts` which haven't changed?

still want to revisit naming at some point, but not intending to break existing map imports here for us or BOPs.

src/components/postcode-search/index.ts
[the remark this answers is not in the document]
DafyddLlyr: Good to know, and thanks for the links! 👍 

Looking forward to getting a bit more hands on with Lit and getting to grips with some of this.

src/components/postcode-search/index.ts
```diff
@@ -33,7 +39,8 @@ export class PostcodeSearch extends LitElement {
   private _showPostcodeError: boolean = false;
 
   _onInputChange(e: any) {
-    // Validate and set Postcode
+    // validate and set postcode
+    //   <input /> uses Lit ".value" syntax to set property, whereas "value" would set attribute
```
DafyddLlyr, line 43: 👍

jessicamcinchak:
[Switch to GOV UK accessible-autocomplete component to fix final audit issues](https://trello.com/c/xy0PoaJT/1839-switch-to-gov-uk-accessible-autocomplete-component-to-fix-final-audit-issues)

vite.config.ts
DafyddLlyr: Yeah from what I can tell this shouldn't break any external calls to the library. The name of the entire package in NPM will be consistent, as will the name used in `index.ts` which is what's used in PlanX, BOPS etc.

We can always give BOPS a heads up and ask them to pin their version of the package until we test if we want to be super cautious. I think we can test this very quickly/easily once deployed though right?

DafyddLlyr (approved):

PR #95, opened 2021-12-10
fix: A11y improvements

DafyddLlyr:
Addresses issues raised in the following two tickets - 
 - https://trello.com/c/NqjvZifP/1666-non-descriptive-links-and-buttons
 - https://trello.com/c/luSUt9gd/1680-button-target-size

**Fixes**
- Ensure zoom and reset buttons are 44px x 44px
- Apply missing `pointer: cursor` to buttons
  - Presumably the way that the drag and move pointers are set are causing this issue
- Apply `aria-label` attribute to controls

image

### DIFF
--- a/src/my-map.ts
+++ b/src/my-map.ts
@@ -58,7 +58,7 @@ export class MyMap extends LitElement {
       cursor: pointer;
       min-width: 44px;
       min-height: 44px;
-      font-size: 1.5rem;
+      font-size: 1.75rem;
     }
     .ol-control button:hover {
       background-color: rgba(44, 44, 44, 0.85) !important;
@@ -248,7 +248,7 @@ export class MyMap extends LitElement {
 
     // Apply aria-labels to OL Controls for accessibility
     const olControls: NodeListOf<HTMLButtonElement> | undefined =
-      this.shadowRoot?.querySelectorAll(".ol-control button");
+      this.renderRoot?.querySelectorAll(".ol-control button");
     olControls?.forEach((node) =>
       node.setAttribute("aria-label", node.getAttribute("title") || "")
     );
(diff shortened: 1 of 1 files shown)


src/my-map.ts
```diff
@@ -242,6 +246,13 @@ export class MyMap extends LitElement {
       map.addControl(ResetControl);
     }
 
+    // Apply aria-labels to OL Controls for accessibility
+    const olControls: NodeListOf<HTMLButtonElement> | undefined =
+      this.shadowRoot?.querySelectorAll(".ol-control button");
```
jessicamcinchak, line 251: Lit nit: I think within `firstUpdated()`, it's more standard/comprehensive to use `renderRoot` rather than `shadowRoot`! The `renderRoot` will be the `shadowRoot` or an element, and still has access to the `.querySelectorAll()` method.

Docs here for reference https://lit.dev/docs/components/shadow-dom/#accessing-nodes-in-the-shadow-dom

src/my-map.ts
```diff
@@ -55,12 +55,16 @@ export class MyMap extends LitElement {
     .ol-control button {
       border-radius: 0 !important;
       background-color: #2c2c2c !important;
+      cursor: pointer;
+      min-width: 44px;
+      min-height: 44px;
+      font-size: 1.5rem;
```
jessicamcinchak, line 61: nit: I actually really like the bigger buttons, but I personally think the font-size-to-padding ratio feels a bit off now? maybe bump font-size up a bit to `1.75rem` ? 

for comparison:
Screenshot from 2021-12-10 16-52-53 (1.5 on the left, and 1.75 on the right) Screenshot from 2021-12-10 16-50-53

jessicamcinchak (approved):
thanks for these improvements!! two minor comments, but happy for this to merge whenever

PR #97, opened 2021-12-11
fix: add disconnectedCallback() lifecycle method to unmount map

### LINKED ISSUE #96: Unmount map
Calling `map.dispose()` when the custom element unmounts might help making sure stale event handlers don't stick around. This could be implemented by setting `this.map = map;` after https://github.com/theopensystemslab/map/blob/main/src/my-map.ts#L179, and then later after https://github.com/theopensystemslab/map/blob/main/src/my-map.ts#L437, adding the following method on the class instance:

```ts
  disconnectedCallback() {
    this.map?.dispose();
  }
```

jessicamcinchak:
fixes #96 

:crossed_fingers: this might/should fix planx bugs described here https://github.com/theopensystemslab/planx-new/pull/704

https://lit.dev/docs/components/lifecycle/#content

### DIFF
--- a/src/my-map.ts
+++ b/src/my-map.ts
@@ -164,6 +164,11 @@ export class MyMap extends LitElement {
   @state()
   map?: Map;
 
+  // called when element is created
+  constructor() {
+    super();
+  }
+
   // runs after the initial render
   firstUpdated() {
     const target = this.renderRoot.querySelector("#map") as HTMLElement;
@@ -209,7 +214,7 @@ export class MyMap extends LitElement {
     });
 
     this.map = map;
-    this.performUpdate();
+    super.performUpdate();
 
     // make configurable interactions available
     const draw = configureDraw(this.drawPointer);
@@ -446,6 +451,7 @@ export class MyMap extends LitElement {
 
   // unmount the map
   disconnectedCallback() {
+    super.disconnectedCallback();
     this.map?.dispose();
   }
 
--- a/src/my-map.ts
+++ b/src/my-map.ts
@@ -1,5 +1,5 @@
 import { css, html, LitElement } from "lit";
-import { customElement, property } from "lit/decorators.js";
+import { customElement, property, state } from "lit/decorators.js";
 import { Control, defaults as defaultControls } from "ol/control";
 import { GeoJSON } from "ol/format";
 import { defaults as defaultInteractions } from "ol/interaction";
@@ -160,9 +160,13 @@ export class MyMap extends LitElement {
   @property({ type: Boolean })
   useScaleBarStyle = false;
 
+  // internal reactive state
+  @state()
+  map?: Map;
+
   // runs after the initial render
   firstUpdated() {
-    const target = this.shadowRoot?.querySelector("#map") as HTMLElement;
+    const target = this.renderRoot.querySelector("#map") as HTMLElement;
 
     const useVectorTiles =
       !this.disableVectorTiles && Boolean(this.osVectorTilesApiKey);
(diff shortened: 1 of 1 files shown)


src/my-map.ts
```diff
@@ -204,6 +208,9 @@ export class MyMap extends LitElement {
       }),
     });
 
+    this.map = map;
+    this.performUpdate();
```
jessicamcinchak, line 212: `this.performUpdate()` feels a bit odd/wrong, but clears up this Lit console warning introduced when setting this.map

Screenshot from 2021-12-11 12-05-02

maybe better approach, or just unnecessary? map behaves as expected with or without this explicit update it seems

src/my-map.ts
```diff
@@ -204,6 +213,9 @@ export class MyMap extends LitElement {
       }),
     });
 
+    this.map = map;
+    super.performUpdate();
```
jessicamcinchak, line 217: calling `performUpdate()` here feels a bit odd, but clears up this Lit console warning introduced when setting this.map - any ideas for better approach? docs seem spotty on this topic

Screenshot from 2021-12-11 12-05-02

jessicamcinchak:
[About the property map returned two unrelated polygons](https://trello.com/c/6nVeR3jp/1619-about-the-property-map-returned-two-unrelated-polygons)

jessicamcinchak:
[Super zoomed out map on 'find property'](https://trello.com/c/YrIzKiXz/1569-super-zoomed-out-map-on-find-property)

jessicamcinchak:
[Map remembers a previously searched property](https://trello.com/c/vMr17Jdy/1538-map-remembers-a-previously-searched-property)

src/my-map.ts
peterszerzo: Would it help to just put `this.map` on the class instance as opposed to setting it as a reactive internal state with the `@state` decorator? The lit docs seem to not mention this as an option but I have a feeling the way `this.map` is used does not need any kind of reactivity (more like a `useRef` than a `useState` to borrow from React lingo).

src/my-map.ts
```diff
@@ -160,9 +160,18 @@ export class MyMap extends LitElement {
   @property({ type: Boolean })
   useScaleBarStyle = false;
 
+  // internal reactive state
+  @state()
```
peterszerzo, line 164: What I mean by the comment on the `performUpdate` - is this line here necessary?

src/my-map.ts
```diff
@@ -230,6 +242,8 @@ export class MyMap extends LitElement {
       }
     };
 
+    // this is an internal event listener, so doesn't need to be removed later
+    // ref https://lit.dev/docs/components/lifecycle/#disconnectedcallback
```
peterszerzo, line 246: Are you sure you get away with not having to clean this up if you didn't use their `render()` method with the `html` tagged template literal?

src/my-map.ts
jessicamcinchak: oh that's clever - i had definitely been following doc patterns and assuming every property should use a decorator, but this works (meaning, I can remove `performUpdate()` without prompting any lifecycle warnings)

src/my-map.ts
jessicamcinchak: removed now!

src/my-map.ts
jessicamcinchak: mostly sure :upside_down_face: the button & it's attached listener seem to be fully encapsulated in the map's shadow DOM on inspection, and the docs say "By default, a bubbling custom event fired inside shadow DOM will stop bubbling when it reaches the shadow root." so I _think_ that means it can be considered "internal" & shouldn't prevent the map from unmounting? 

https://lit.dev/docs/v1/components/events/#add-event-listeners-after-first-paint

peterszerzo: Oh right, that makes sense!

DafyddLlyr (approved):
👍

PR #102, opened 2021-12-20
fix: in draw mode, reset button should dispatch areaChange event with value 0

jessicamcinchak:


jessicamcinchak:
[[super minor] boundary area doesn't reset when the boundary is reset](https://trello.com/c/eltFtPDu/1717-super-minor-boundary-area-doesnt-reset-when-the-boundary-is-reset)

src/my-map.ts
```diff
@@ -239,6 +239,10 @@ export class MyMap extends LitElement {
 
       if (this.drawMode) {
         drawingSource.clear();
+
+        this.dispatch("geojsonChange", {});
+        this.dispatch("areaChange", `0 ${this.areaUnit}`);
```
gunar, line 244: The function `formatArea` produces a slightly different result for `areaUnit === 'm2'`:

image


Wondering if we should extract just the unit formatting logic so that we it can be used both there and here.

gunar (approved):

src/my-map.ts
jessicamcinchak: good eye, thanks for spotting this! fixed it the lazy way for now, but will revisit original helper method logic soon!

PR #110, opened 2022-01-13
fix: add prop to set unique id on element & clear the featureSource unless `clickFeatures` is enabled

jessicamcinchak:
**working solution:**
- add an `id` property so we can set a custom id on `<my-map />` element, defaults to "map" as before
- clear the featureSource upfront when only `showFeaturesAtPoint` is enabled, which should prevent prior query results from persisting in the layer. if `showFeaturesAtPoint` && `clickFeatures` are enabled (don't know of any prod uses of this yet), then we don't clear the source as before to allow for selecting/deselecting features on click
- clear the drawingSource upfront for _both_ cases: when we want to start drawing fresh AND when we want to load a previously drawn polygon into the drawingSource to modify. previously, we only cleared it in the case of starting a fresh drawing
 
**dugging notes:**
update `index.html` to something like this:
```html
<body>
  <div style="display: flex; flex-direction: column;">
    <div style="margin-bottom: 1em">
      <my-map showFeaturesAtPoint latitude="51.507351" longitude="-0.127758" id="map-1" />
    </div>
    <div>
      <my-map showFeaturesAtPoint latitude="51.506351" longitude="-0.124758" id="map-2" />
    </div>
  </div>
</body>
```

in short: when rendering the component multiple times on a page with different props, each instance reflects the layer data of the other -  some props sync unexpectedly (ie `showFeaturesAtPoint`, `drawMode`), and others do not (`featureColor`, etc).

expected behavior is for each instance of the component to exclusively reflects its' own props & it's own layer data based on those props (eg only the feature that intersects with its' unique lat/lng). 

this can be reproduced in PlanX by selecting an address, going 'back' to change the address, and then seeing two highlighted features on the FindProperty map

things i learned: 
- "layers can only be added to a single map" ([openlayers docs](https://github.com/openlayers/openlayers/blob/main/changelog/upgrade-notes.md#layers-can-only-be-added-to-a-single-map)) 
  - so even though these index.html map instances have unique ids and different props, `outlineLayer`, `drawingLayer`, etc are still shared/synced between them. Here's related threads/issues: https://github.com/openlayers/openlayers/issues/10764#issuecomment-596259650, https://stackoverflow.com/questions/63267102/how-can-i-assign-a-unique-identifier-to-openlayers-layers
- since we're experiencing the synced layer data in planx, that indicates that we might have multiple maps rendering on top of one another, or while the map itself re-draws after receiving new props, it's data layer hasn't actually cleared - these working solutions should help that :crossed_fingers:

### DIFF
--- a/index.html
+++ b/index.html
@@ -14,16 +14,9 @@
     />
   </head>
   <body>
-    <div style="display: flex; flex-direction: column;">
-      <div style="margin: 1em">
-        <my-map showFeaturesAtPoint latitude="51.507351" longitude="-0.127758" id="map-1" />
-      </div>
-      <div style="margin: 1em">
-        <my-map showFeaturesAtPoint featureColor="#ff00ff" drawMode drawPointer="dot" latitude="51.506351" longitude="-0.124758" id="map-2" />
-      </div>
-    </div>
+    <my-map zoom="18" maxZoom="23" drawMode drawPointer="dot" id="example-map" />
 
-    <!-- <script>
+    <script>
       const map = document.querySelector("my-map");
 
       map.addEventListener("ready", (event) => {
@@ -50,6 +43,6 @@
       map.addEventListener("geojsonDataArea", ({ detail: geojsonDataArea }) => {
         console.debug({ geojsonDataArea });
       });
-    </script> -->
+    </script>
   </body>
 </html>
--- a/index.html
+++ b/index.html
@@ -14,9 +14,16 @@
     />
   </head>
   <body>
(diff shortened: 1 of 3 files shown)


jessicamcinchak:
[Super zoomed out map on 'find property'](https://trello.com/c/YrIzKiXz/1569-super-zoomed-out-map-on-find-property)

jessicamcinchak:
[Map remembers a previously searched property](https://trello.com/c/vMr17Jdy/1538-map-remembers-a-previously-searched-property)

DafyddLlyr (approved):
All the changes here look good and make sense to me 👍

PR #112, opened 2022-02-04
fix: display snaps on first render if zoom within threshold

jessicamcinchak:
Previosly, we'd only render snaps after the map moves (zoom, pan, etc) - this change makes snaps visible on the first render before any interactions if other conditions are met (`drawMode` is enabled, `zoom` is greater than or equal to 20)

### DIFF
--- a/src/my-map.ts
+++ b/src/my-map.ts
@@ -381,23 +381,35 @@ export class MyMap extends LitElement {
       });
     }
 
-    // show snapping points when in drawMode, with vector tile basemap enabled, and at zoom > 20
+    // show snapping points when in drawMode, with vector tile basemap enabled, and at qualifying zoom
     if (
       this.drawMode &&
       Boolean(this.osVectorTilesApiKey) &&
       !this.disableVectorTiles
     ) {
-      map.addLayer(pointsLayer);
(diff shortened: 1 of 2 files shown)


jessicamcinchak:
[Increase default zoom on map so snaps are immediately visible](https://trello.com/c/Z0wZoCKJ/1758-increase-default-zoom-on-map-so-snaps-are-immediately-visible)

src/my-map.ts
```diff
@@ -381,23 +381,35 @@ export class MyMap extends LitElement {
...
 
-    // show snapping points when in drawMode, with vector tile basemap enabled, and at zoom > 20
+    // show snapping points when in drawMode, with vector tile basemap enabled, and at qualifying zoom
     if (
       this.drawMode &&
       Boolean(this.osVectorTilesApiKey) &&
       !this.disableVectorTiles
     ) {
+      // define zoom threshold for showing snaps (not @property yet because computationally expensive!)
+      const snapsZoom: number = 20;
```
DafyddLlyr, line 391: I agree that making this accessible would be a bad idea!

DafyddLlyr (approved):
Great 👍

PR #114, opened 2022-02-25
`pnpm i` to bring lockfile up to date with package.json

johnrees:
I just ran `pnpm i` so that the pnpm-lock.yaml is up to date with the deps specified in package.json

Screenshot 2022-02-24 at 17 40 45

not sure what happened with the commit name and description but if it's a squash and merge I suppose they don't matter


jessicamcinchak (approved):
thanks for catching this!

PR #120, opened 2022-03-22
feat: address-autocomplete supports a default value

jessicamcinchak:
ability to set an `initialAddress` prop on the web component, which is then passed to accessible-autocomplete's `defaultValue` option: https://github.com/alphagov/accessible-autocomplete. 

this will enable us to show the prior selected address when "going back" or "changing" your site address in planx. 

since our `initialAddress` will only ever be one that was selected prior, we can assume that we already have the full address record stored in the passport/state from a prior event dispatch, and don't need to fire a new one on initial render.

small bonus of improved options sorting in the dropdown - before address "103..." would show up before "47..." etc.

### DIFF
--- a/src/components/address-autocomplete/index.ts
+++ b/src/components/address-autocomplete/index.ts
@@ -25,6 +25,9 @@ export class AddressAutocomplete extends LitElement {
   label = "Select an address";
 
   @property({ type: String })
+  initialAddress = "";
+
+  @property({ type: String })
   osPlacesApiKey = import.meta.env.VITE_APP_OS_PLACES_API_KEY || "";
 
   // internal reactive state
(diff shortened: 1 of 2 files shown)


src/components/address-autocomplete/index.ts
```diff
@@ -124,7 +128,8 @@ export class AddressAutocomplete extends LitElement {
               );
             });
 
-          this._options.sort((a, b) => a.localeCompare(b));
+          const collator = new Intl.Collator([], { numeric: true });
```
gunar, line 131: TIL https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator

gunar (approved):

src/components/address-autocomplete/index.ts
jessicamcinchak: me too :slightly_smiling_face:

PR #124, opened 2022-03-23
feat: Allow autocomplete to be styled from parent

DafyddLlyr:
This PR allows certain properties of the `address-autocomplete` component to be styled from a parent. The properties that have been given variables should marry up with any that we need to match the PlanX style.

Lit Docs - https://lit.dev/docs/components/styles/#theming

**Example with Lambeth primary colour**

image

**To Do**
- [ ] Document these new properties with more than just a comment in the code

### DIFF
--- a/index.html
+++ b/index.html
@@ -12,6 +12,22 @@
       href="https://fonts.googleapis.com/css2?family=Source+Sans+Pro:wght@300;400;600&display=swap"
       rel="stylesheet"
     />
+    <!-- Examples of available style options for address-autocomplete -->
+    <!-- <style>
+      address-autocomplete {
+        --autocomplete__input__padding: 6px 12px 7px 12px;
+        --autocomplete__input__font-size: 15px;
+        --autocomplete__input__height: 50px;
+        --autocomplete__dropdown-arrow-down__top: 16px;
+        --autocomplete__dropdown-arrow-down__z-index: 2;
+        --autocomplete__option__font-size: 15px;
+        --autocomplete__option__padding: 6px 12px 7px 12px;
+        --autocomplete__menu__max-height: 336px;
+        --autocomplete__option__border-bottom: solid 1px grey;
+        --autocomplete__option__hover-border-color: pink;
+        --autocomplete__option__hover-background-color: pink;
+      }
+    </style> -->
   </head>
   <body>
     <div style="display:flex;flex-direction:column;">
--- a/index.html
+++ b/index.html
@@ -13,7 +13,7 @@
       rel="stylesheet"
     />
     <!-- Examples of available style options for address-autocomplete -->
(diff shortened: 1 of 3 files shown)


src/components/address-autocomplete/index.ts
```diff
@@ -170,7 +171,11 @@ export class AddressAutocomplete extends LitElement {
             href="https://cdn.jsdelivr.net/npm/accessible-autocomplete@2.0.4/dist/accessible-autocomplete.min.css"
           />
           <label class="govuk-label" htmlFor=${this.id}> ${this.label} </label>
-          <div id="${this.id}-container" role="status"></div>`;
+          <div
+            id="${this.id}-container"
+            role="status"
+            spellcheck="false"
+          ></div>`;
```
jessicamcinchak, line 178: :+1: thanks for catching these extra props!

src/components/address-autocomplete/styles.scss
```diff
@@ -1,6 +1,46 @@
 @import "node_modules/govuk-frontend/govuk/all";
 // @import "accessible-autocomplete";
 
+:host {
+  $font-size: var(--autocomplete__input__font-size, 19px);
+  $input-height: var(--autocomplete__input__height, 35px);
```
jessicamcinchak, line 6: maybe option to pass `$font-family` here too? i think this could default to `"GDS Transport", arial, sans-serif"` as it renders now, but we'll probably want it to inherit Inter in planx? (otherwise i think it would become the new stray rounded-corners that would really haunt alastair lol :upside_down_face:)

i have a weird little hack below for the `ul, li` elements you should feel free to remove too - i realize that font does not actually match the autocomplete label currently!

index.html
```diff
@@ -12,6 +12,22 @@
...
+        --autocomplete__dropdown-arrow-down__top: 16px;
+        --autocomplete__dropdown-arrow-down__z-index: 2;
+        --autocomplete__option__font-size: 15px;
+        --autocomplete__option__padding: 6px 12px 7px 12px;
+        --autocomplete__menu__max-height: 336px;
+        --autocomplete__option__border-bottom: solid 1px grey;
+        --autocomplete__option__hover-border-color: pink;
+        --autocomplete__option__hover-background-color: pink;
+      }
+    </style> -->
```
jessicamcinchak, line 30: cool to see theming in action, this all works smoothly for me locally :hibiscus:

jessicamcinchak (approved):
thank you, this is great! one minor suggestion, but excited to try this in planx asap :rocket:

src/components/address-autocomplete/styles.scss
DafyddLlyr: Good catch! ⚾

PR #128, opened 2022-03-24
feat: Improve style of autocomplete

DafyddLlyr:
A few little fixes and improvements.

 - Down arrow is visible on white background, and is now clickable (only to open, as original component works)
   - Relates to this issue https://github.com/alphagov/accessible-autocomplete/issues/351
 - New "light" down arrow option which is a little nicer looking 💅 
 - Style doesn't break when focus and hover interact (see below)
   - There's still a `:focus-visible` ring, see issue here - https://github.com/alphagov/accessible-autocomplete/issues/73  

chrome-capture (12)

### DIFF
--- a/src/components/address-autocomplete/index.ts
+++ b/src/components/address-autocomplete/index.ts
@@ -9,6 +9,8 @@ type Address = {
   LPI: any;
 };
 
+type ArrowStyleEnum = "default" | "light";
+
 @customElement("address-autocomplete")
 export class AddressAutocomplete extends LitElement {
   // ref https://github.com/e111077/vite-lit-element-ts-sass/issues/3
@@ -31,7 +33,7 @@ export class AddressAutocomplete extends LitElement {
   osPlacesApiKey = import.meta.env.VITE_APP_OS_PLACES_API_KEY || "";
 
   @property({ type: String })
-  arrowStyle = "default";
+  arrowStyle: ArrowStyleEnum = "default";
 
   // internal reactive state
   @state()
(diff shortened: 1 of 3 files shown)


src/components/address-autocomplete/index.ts
```diff
@@ -30,6 +30,9 @@ export class AddressAutocomplete extends LitElement {
   @property({ type: String })
   osPlacesApiKey = import.meta.env.VITE_APP_OS_PLACES_API_KEY || "";
 
+  @property({ type: String })
+  arrowStyle = "default";
```
jessicamcinchak, line 34: i've been defining typescript types in lieu of Lit offering an enum property type, that might make sense here too? usually just leads to clearer error messages if debugging, nicer vscode prefill suggestions, etc

here's an example from the map properties: 
```
type DrawPointerEnum = "crosshair" | "dot";

@property({ type: String })
drawPointer: DrawPointerEnum = "crosshair";
```

src/components/address-autocomplete/index.ts
```diff
@@ -57,6 +60,10 @@ export class AddressAutocomplete extends LitElement {
     super.disconnectedCallback();
   }
 
+  getLightDropdownArrow() {
+    return '<svg class="autocomplete__dropdown-arrow-down" style="height: 17px;" viewBox="0 0 512 512" ><path d="M256,298.3L256,298.3L256,298.3l174.2-167.2c4.3-4.2,11.4-4.1,15.8,0.2l30.6,29.9c4.4,4.3,4.5,11.3,0.2,15.5L264.1,380.9  c-2.2,2.2-5.2,3.2-8.1,3c-3,0.1-5.9-0.9-8.1-3L35.2,176.7c-4.3-4.2-4.2-11.2,0.2-15.5L66,131.3c4.4-4.3,11.5-4.4,15.8-0.2L256,298.3  z"/></svg>';
+  }
```
jessicamcinchak, line 65: :+1: nice + simple

jessicamcinchak (approved):
visually subtle, but so much more usable - thanks!!

PR #130, opened 2022-03-28
feat: Enable static or responsive label on autocomplete

DafyddLlyr:
Realised during Friday's show and tell that the label font size is 19px on PlanX when it should be 18px!

By default, the `govuk-label` class takes into account the following breakpoints - https://github.com/alphagov/govuk-frontend/blob/v4.0.0/src/govuk/settings/_media-queries.scss#L10-L14

This means it's not quite as simple as overwriting the font-size and setting a default, as this would have the side effect of breaking the responsiveness if anybody else wanted to use the component.

### DIFF
--- a/src/components/address-autocomplete/index.ts
+++ b/src/components/address-autocomplete/index.ts
@@ -168,7 +168,7 @@ export class AddressAutocomplete extends LitElement {
       .catch((error) => console.log(error));
   }
 
-  getLabelClasses() {
+  _getLabelClasses() {
     let styles = "govuk-label";
     if (this.labelStyle === "static") {
       styles += " govuk-label--static";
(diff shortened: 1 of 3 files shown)


src/components/address-autocomplete/index.ts
```diff
@@ -164,6 +168,14 @@ export class AddressAutocomplete extends LitElement {
       .catch((error) => console.log(error));
   }
 
+  getLabelClasses() {
```
jessicamcinchak, line 171: nit: can you please update this to be `_getLabelClasses()` - trying to follow common Lit pattern of using for the `_` prefix for all non-Lit-lifecycle methods

src/components/address-autocomplete/index.ts
```diff
@@ -186,7 +198,9 @@ export class AddressAutocomplete extends LitElement {
             rel="stylesheet"
             href="https://cdn.jsdelivr.net/npm/accessible-autocomplete@2.0.4/dist/accessible-autocomplete.min.css"
           />
-          <label class="govuk-label" htmlFor=${this.id}>${this.label}</label>
+          <label class=${this.getLabelClasses()} htmlFor=${this.id}
```
jessicamcinchak, line 201: make sure to rebase to not accidently overwrite with `htmlFor` again :slightly_smiling_face:

jessicamcinchak (approved):
good eye, good solution for our use case! thanks for this!

src/components/address-autocomplete/index.ts
DafyddLlyr: Very good point...!

PR #136, opened 2022-03-29
fix: Ensure error container is always in DOM (autocomplete)

DafyddLlyr:


jessicamcinchak (approved):
nice solution, each render role="status" as expected for me!

PR #138, opened 2022-03-30
fix: postcode-search improvements

jessicamcinchak:
- Adjust error message according to new `govuk-frontend` release [change note](https://github.com/alphagov/govuk-frontend/blob/main/CHANGELOG.md#update-the-html-for-error-messages)
- Dispatch an event on input change, rather than only when the postcode is validated - otherwise an integration in planx doesn't know when to (re)hide the address-autocomplete when a valid postcode is changed
- Added basic set of custom style properties ([Lit docs](https://lit.dev/docs/v1/components/styles/#configurable)) we'd likely want to configure/override in planx, following autocomplete example, but nothing too fancy or complex
- Use `for` not `htmlFor` to link label & input

### DIFF
--- a/src/components/postcode-search/index.ts
+++ b/src/components/postcode-search/index.ts
@@ -94,13 +94,11 @@ export class PostcodeSearch extends LitElement {
   _makeLabel() {
     return this.onlyQuestionOnPage
       ? html`<h1 class="govuk-label-wrapper">
-          <label class="govuk-label govuk-label--l" htmlFor=${this.id}>
+          <label class="govuk-label govuk-label--l" for=${this.id}>
             ${this.label}
           </label>
         </h1>`
-      : html`<label class="govuk-label" htmlFor=${this.id}
-          >${this.label}</label
-        >`;
+      : html`<label class="govuk-label" for=${this.id}>${this.label}</label>`;
   }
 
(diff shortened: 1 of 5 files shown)


src/components/postcode-search/index.ts
```diff
@@ -75,6 +75,8 @@ export class PostcodeSearch extends LitElement {
   _showError() {
     const errorEl: HTMLElement | null | undefined =
       this.shadowRoot?.querySelector(`#${this.errorId}`);
+
+    // display "none" ensures always present in DOM, which means role="status" will work for screenreaders
```
gunar, line 79: :+1:

src/components/postcode-search/styles.scss
```diff
@@ -1,2 +1,28 @@
...
+  .govuk-error-message {
+    font-family: $font-family;
+    font-size: $font-size;
+  }
+
+  .govuk-input {
+    font-family: $font-family;
+    font-size: $input-font-size;
+    height: $input-height;
+    padding: var(--postcode__input__padding, 5px 34px 5px 5px);
```
gunar, line 26: CSS3 is so cool

gunar (approved):

PR #139, opened 2022-03-30
feat: unit tests

### LINKED ISSUE #52: Add tests
https://lit.dev/docs/tools/testing/

https://open-wc.org/docs/testing/helpers/

examples:
with Karma: https://www.thisdot.co/blog/testing-web-components-with-karma-and-typescript
with Jest: https://www.ninkovic.dev/blog/2020/testing-web-components-with-jest-and-lit-element

jessicamcinchak:
trying out this stack! 
https://vitest.dev/
https://www.npmjs.com/package/happy-dom
https://github.com/vitest-dev/vitest/tree/main/examples/lit

initially tried `@open-wc/testing` with `jest` first which spiraled into a lot of tedious configs (ts, jest, babel etc) and wasn't playing so nice with vite, but that then led me to `vitest` which was very simple to set up (unified config! runs fast!) & has good selection of example repos - including one for Lit - so let's see how it feels!

plus a simple github action to automatically run these tests in PRs against `main`, starting with this one :checkered_flag: 
Screenshot from 2022-03-31 17-31-50

closes #52

in the future, maybe we additionally add cyprus for end to end tests, which i think could especially make sense for map interactions? these vitest + cyprus videos (part [1](https://www.youtube.com/watch?v=D-4VBU2wkSk ) & [2](https://www.youtube.com/watch?v=Hgfd0SX6tgs)) seem pretty good, but i haven't watched them through yet


src/components/address-autocomplete/main.test.ts
```diff
@@ -0,0 +1,68 @@
...
+
+declare global {
+  interface Window extends IWindow {}
+}
+
+describe("AddressAutocomplete on initial render with valid postcode", async () => {
+  beforeEach(async () => {
+    document.body.innerHTML = `<address-autocomplete id="autocomplete-vitest" postcode="SE5 0HU" osPlacesApiKey=${
+      import.meta.env.VITE_APP_OS_PLACES_API_KEY
+    } />`;
```
jessicamcinchak, line 16: feels a bit strange to pass API keys in tests, but it's a required prop to render the component - should look into how to mock related fetch response??

make sure your local secrets are set in `env.local` to be picked up by `pnpm test`; `env.development.local` will only be picked up by `pnpm dev`

for the github action, i've added our API key to repository secrets

src/components/postcode-search/main.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    // input is empty on render
+    const input = getShadowRootEl("postcode-search", "input");
+    let inputValue = (input as HTMLInputElement)!.value;
+    expect(inputValue).toEqual("");
+
+    // set input value, expect event to dispatch
+    inputValue = "S";
+    expect(inputValue).toEqual("S");
+    expect(spyPostcodeChange).toHaveBeenCalled();
```
jessicamcinchak, line 125: mimicing some interaction (eg typing) & spying if an event dispatch has been called has been finicky to figure out - all `expect` statements here are passing until the final one. my suspicions are perhaps that this is connected to Lit's `.value` syntax, but not sure yet?

the button example here was helpful to get started: https://github.com/vitest-dev/vitest/blob/main/examples/lit/test/basic.test.ts#L31-L39

src/components/postcode-search/main.test.ts
```diff
@@ -0,0 +1,127 @@
+import type { IWindow } from "happy-dom";
+import { beforeEach, describe, it, expect, vi } from "vitest";
+
+import { getShadowRoot, getShadowRootEl } from "../../test-utils";
+
+import "./index";
```
jessicamcinchak, line 6: i've written the most tests for the one component not in production, whoops :shrug: lots of room for improved coverage!

src/components/postcode-search/main.test.ts
DafyddLlyr: Yeah I've spent a bit of time this morning trying to get this to pass and get my head around what's happening here. I've got it working, but I'm not 100% sure of the cause of the problem. I think you are correct though - us changing the value does not trigger the event or update the value in Lit's state essentially.

Eventually got it working though! I added `@testing-library/dom` & `@testing-library/user-event` as devDependencies and then updated the code to this - 


```js
it("should dispatch event on input change", async () => {
  const spyPostcodeChange = vi.fn();

  document
    .querySelector("postcode-search")!
    .addEventListener("postcodeChange", spyPostcodeChange);

  // input is empty on render
  const input = getShadowRootEl("postcode-search", "input") as HTMLInputElement;
  expect(input).toBeTruthy();
  expect(input.value).toEqual("");

  // set input value, expect event to dispatch
  await userEvent.type(input, "S")
  expect(input.value).toEqual("S");
  expect(spyPostcodeChange).toHaveBeenCalled();
});
```

src/test-utils.ts
```diff
@@ -0,0 +1,17 @@
...
+  return document.body.querySelector(customEl)?.shadowRoot;
+}
+
+// Helper method to access a specific HTML element within the shadow root of a custom element
+export function getShadowRootEl(
+  customEl: string,
+  el: string
+): Element | null | undefined {
+  return document.body.querySelector(customEl)?.shadowRoot?.querySelector(el);
+}
```
DafyddLlyr, line 12: I think there's likely a clever way using TypeScript generics to avoid passing the element in, and then casting to the element on return. Something to look at in future maybe...! I had something like this in mind, but couldn't get it to work - 

```ts
export function getShadowRootEl<T extends HTMLElement>(
  customEl: string,
  type: T,
): T | null | undefined {
  return document.body.querySelector(customEl)?.shadowRoot?.querySelector(type.tagName) as T;
}
```

And then calling it as - 

```ts
const input = getShadowRootEl("postcode-search", HTMLInputElement);
```

DafyddLlyr (approved):
Embarrassed by how long it's taken me to write this up sorry...! 😬 

Stack looks and feels good, tried writing tests and getting to grips with it and I like it. Totally agree with what you're saying about Cypress for maps and interactions - might be worth testing Cypress here and then seeing how we feel with that vs testcafe in PlanX even?

Anyway - solid work, this adds a lot to the repo and is a great foundation 👏

src/components/postcode-search/main.test.ts
```diff
@@ -0,0 +1,127 @@
...
+      "input"
+    ) as HTMLInputElement;
+    input?.focus();
+    input?.blur();
+
+    const error = getShadowRoot("postcode-search")?.getElementById(
+      "postcode-error-vitest"
+    );
+    expect(error).toBeDefined();
+    expect(error?.innerHTML).toContain("Enter a valid UK postcode");
```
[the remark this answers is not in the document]
jessicamcinchak: good shout on this one - now updated to more accurately assert that style is `"display:none"` before blur and null after

src/test-utils.ts
jessicamcinchak: Yep, definitely interested in revisiting this and building out smarter test-utils here - agree these first-go methods feel a little flimsy!

src/components/postcode-search/main.test.ts
jessicamcinchak: Thanks for spending some time with this one! Your solution also works for me too :clap: will push this up. 

Here's a short list of maybe-useful threads for future reading as we continue to wrap our heads around this!! :bookmark: 
https://github.com/capricorn86/happy-dom/issues/313
https://stackoverflow.com/questions/71770813/why-cant-testing-library-user-event-interact-with-input-elements
https://stackoverflow.com/questions/72783541/vitest-litelement-events

PR #159, opened 2022-07-05
feat: show features without border & ability to add a custom marker

jessicamcinchak:
two styling features to help clarify how we highlight addresses: 
- boolean property `featureBorderNone` to control whether or not to show a border when highlighting an OS Feature that intersects with a point
```<my-map zoom="20" showFeaturesAtPoint featureFill="#fff" featureBorderNone />```
Screenshot from 2022-07-05 12-58-22

- boolean property `showMarker` to show a marker on the map (defaults to center latitude & longitude), with optional number properties `markerLatitude`, `markerLongitude` to specify a custom latitude & longitude to place the marker at. The marker will default to a black dot, but you can pass a custom hex code or CSS color name to `markerColor`.
```<my-map zoom="20" showMarker />```
Screenshot from 2022-07-05 13-50-53

### DIFF
--- a/src/components/my-map/index.ts
+++ b/src/components/my-map/index.ts
@@ -85,6 +85,9 @@ export class MyMap extends LitElement {
   @property({ type: Boolean })
   featureFill = false;
 
+  @property({ type: Boolean })
+  featureBorderNone = false;
+
   @property({ type: Number })
   featureBuffer = 40;
 
@@ -397,7 +400,8 @@ export class MyMap extends LitElement {
 
       const outlineLayer = makeFeatureLayer(
         this.featureColor,
-        this.featureFill
+        this.featureFill,
+        this.featureBorderNone
       );
       map.addLayer(outlineLayer);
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/components/my-map/index.ts [hunk 1]
```diff
@@ -85,9 +87,24 @@ export class MyMap extends LitElement {
   @property({ type: Boolean })
   featureFill = false;
 
+  @property({ type: Boolean })
+  featureBorderNone = false;
```
gunar, line 91: Nit: I tend to favor positive naming schemes (e.g. `featureBorder = true`) as to avoid double negation (e.g. `featureBorderNone = false`).

gunar (approved):

src/components/my-map/index.ts
jessicamcinchak: See https://github.com/theopensystemslab/map/issues/37 - had originally defaulted to positive naming schemes for boolean properties, but then learned overtime that it doesn't play so nice with Lit so now consistently declaring all boolean props `false` by default.

PR #167, opened 2022-07-28
build: update to vite v3

jessicamcinchak:
Now using Vite v3 :tada: 

https://vitejs.dev/guide/migration.html

Notably, our Netlify build container was trying to use Node v12, whereas Vite now requires 14.18+ / 16+, so I've added an `.nvmrc` file.

Otherwise, no breaking changes for us, just smaller build size and even speedier :zap: 

I expect this to automatically close dependabots PR #164 when merged


DafyddLlyr (approved):
🎉

PR #191, opened 2022-08-19
feat: add basic unit tests for map component

jessicamcinchak:
recent dependency upgrades seem to have cleared up the previous openlayers type error that was effecting map unit tests :tada: 

additionally updated all unit tests to use `toBeTruthy` rather than `toBeDefined` as that would still pass for null elements


DafyddLlyr (approved):
Nice!

PR #193, opened 2022-08-25
fix: Ensure snap points always load

DafyddLlyr:
Addresses https://trello.com/c/NZrvqg7d/2055-get-snap-points-on-map-to-always-load

Currently, points are automatically generated on the assumption that vector tiles will be loaded. On slower connections, this means that snap points either do not display or do not fully display. This can be mocked in the browser by throttling your connection speed.

This PR fixes this by waiting for the map `loadend` event before triggering the generation of the snap points. Originally I was listening for the vector source's `tileloadend` event but this fires for each tile load and then regenerates snap points which I imagine is both computationally expensive and could lead to inconsistent results (e.g. all snap points loading in one tile at a time).

Also - some basic OL tests! Couldn't quite get around a few event dispatch issues so I'm manually calling them currently. This would be great to try and remove - possibly something to tackle when/if we add Cypress to the stack.

### DIFF
--- a/src/components/my-map/index.ts
+++ b/src/components/my-map/index.ts
@@ -222,6 +222,9 @@ export class MyMap extends LitElement {
 
     this.map = map;
 
+    // Append to global window for reference in tests
+    window.olMap = import.meta.env.VITEST ? this.map : undefined;
+
     // make configurable interactions available
     const draw = configureDraw(this.drawPointer);
     const modify = configureModify(this.drawPointer);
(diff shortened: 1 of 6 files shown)


DafyddLlyr:
[Get snap points on map to always load ](https://trello.com/c/NZrvqg7d/2055-get-snap-points-on-map-to-always-load)

src/components/my-map/index.ts
```diff
@@ -535,6 +536,9 @@ export class MyMap extends LitElement {
 }
 
 declare global {
+  interface Window {
+    olMap: Map | undefined;
+  }
```
jessicamcinchak, line 541: :heavy_plus_sign:

src/components/my-map/main.test.ts
```diff
@@ -1,20 +1,36 @@
...
+  document.body.innerHTML = mapElement;
+  await window.happyDOM.whenAsyncComplete();
+  window.olMap?.dispatchEvent("loadend");
+};
 
-    await window.happyDOM.whenAsyncComplete();
-  }, 2500);
+test("olMap is added to the global window for tests", async () => {
+  await setupMap(`<my-map id="map-vitest" disableVectorTiles />`);
+  expect(window.olMap).not.toBeUndefined();
```
jessicamcinchak, line 25: could this be `toBeTruthy` instead of `not.toBeUndefined` for readability?

src/components/my-map/main.test.ts
```diff
@@ -30,3 +46,89 @@ describe("MyMap on initial render with OSM basemap", async () => {
...
+  });
+
+  it("should load snap points if the initial zoom is within range", async () => {
+    await setupMap(`<my-map 
+      id="map-vitest" 
+      zoom=${ZOOM_WITHIN_RANGE} 
+      drawMode 
+      osVectorTileApiKey=${process.env.VITE_APP_OS_VECTOR_TILES_API_KEY}
+    />`);
+    expect(getSnapSpy).toHaveBeenCalled();
```
jessicamcinchak, line 85: can we make this a bit stronger and specify `.toHaveBeenCalledOnce`? as computing & loading snaps is a bit expensive, feels useful to monitor we aren't making unnecessary requests!

jessicamcinchak (approved):
This works great for me - snaps load a lot more consistently on a slow connection! 

Thanks for tests too - there's some great foundational stuff there in addition to the snap-specific logic. A few minor comments but no blockers for me!

src/components/my-map/main.test.ts
DafyddLlyr: This is a great suggestion 💪

PR #198, opened 2022-10-05
feat: add boolean prop `showNorthArrow` to toggle on custom control

jessicamcinchak:
changes:
- adds boolean prop `showNorthArrow` which, when true, will display a custom control image; intentionally _not_ using OL's built in rotation control here because we don't want it to display as a button, this is an informational control only. the rotation interaction is still disabled.
- adds icons subdirectory and source notes
- renames `scale-line.ts` to `controls.ts`
Screenshot from 2022-10-05 09-31-24


jessicamcinchak:
[Add north arrow to site map](https://trello.com/c/xelgKeRM/2123-add-north-arrow-to-site-map)

src/components/my-map/controls.ts
```diff
@@ -0,0 +1,25 @@
...
+  const image = document.createElement("img");
+  image.src = northArrowIcon;
+  image.title = "North";
+
+  const element = document.createElement("div");
+  element.className = "north-arrow-control ol-unselectable ol-control";
+  element.appendChild(image);
+
+  return new Control({ element: element });
+}
```
DafyddLlyr, line 25: Something to consider / another approach here - 

When the map loads, you might notice that the [Rotate control](https://openlayers.org/en/latest/apidoc/module-ol_control_Rotate-Rotate.html) actually flashes up for a second. This is because it's part of the `defaultControls` that we're importing. It has `autoHide = true` which means it doesn't display when rotation is at 0.

We could set `autoHide = false`, and modify the inbuilt Rotate control.

Very possible you've already tried the above I appreciate!

DafyddLlyr (changes requested):
One minor thing to look at - happy to go with the approach here if there's a reason to use it over the default `Rotate` (e.g. styling was a pain), but we should set `rotate = false` in the `defaultControls` in `my-map/index.ts` if going with a custom control 👍

src/components/my-map/controls.ts
jessicamcinchak: I did fiddle with this inbuilt `autoHide = false` approach for a very long time yesterday, but eventually decided to pivot to this custom control because overriding built in "button" (even when it didn't have an onClick function) to be treated as an "image" was difficult and styling was tedious. I feel more confident that we won't introduce any confusing interactions with this image as it should seem less "clickable". 

Thanks or spotting lack of `rotate: false` in `defaultControls` though, that's updated now!

DafyddLlyr: Perfect - thought that might have been the case here but just wanted to check 🙂

PR #203, opened 2022-10-10
build: upgrade testing dependencies

jessicamcinchak:
Tests succeed here in their usual ~30sec, but CI fails to recognize the step is done and keeps running until eventually timing out: 
Screenshot from 2022-10-10 14-48-53

I've tried a bunch of different suggestions from these threads with no luck:
- https://github.com/vitest-dev/vitest/issues/1971, addressed in [release v0.23.1](https://github.com/vitest-dev/vitest/releases/tag/v0.23.1)
- https://github.com/vitest-dev/vitest/issues/1288
- https://github.com/vitest-dev/vitest/issues/2008

Suggestions welcome!!

### DIFF
--- a/.github/workflows/pnpm-test.yml
+++ b/.github/workflows/pnpm-test.yml
@@ -23,7 +23,7 @@ jobs:
           cache: "pnpm"
           cache-dependency-path: "**/pnpm-lock.yaml"
       - run: pnpm install --no-frozen-lockfile
-      - run: pnpm run test:ci -- --coverage --watch false
+      - run: pnpm run test:ci
         env:
           VITE_APP_OS_PLACES_API_KEY: ${{ secrets.OS_API_KEY }}
           VITE_APP_OS_VECTOR_TILES_API_KEY: ${{ secrets.OS_API_KEY }}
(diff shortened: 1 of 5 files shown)


DafyddLlyr:
@jessicamcinchak Did a bit of deep dive on this last night without much success. I can recreate exactly what you're describing above - all tests pass but the test runner doesn't quit.

Jest has a `--detectOpenHandles` flag but I can't see anything similar in the vitest docs that would allow us to identify any processes which are still running.

One thing of note - the regression appears to have been introduced in [vitest v0.22.0](https://github.com/vitest-dev/vitest/releases/tag/v0.22.0) - any higher version seems to have this issue, but the previous version works fine (v0.21.1). One pragmatic solution (for now) could be to upgrade to this version.

I also tried narrowing down suites of tests to see if there was a particular one which was causing this. The issue pretty much seems to be with all tests (so maybe something related to `happyDom().whenAsyncComplete()`...?). A simple `expect(true).toBe(true)` passes and the test runner quits.

Going to keep looking at it but wanted to share findings so far - looking through happyDom docs and releases now!

PR #209, opened 2022-10-14
feat: add `resetControlImage` prop for testing custom icons

jessicamcinchak:
adds `resetControlImage` ENUM property to specify a custom icon to show on the reset control button. Defaults to our current unicode arrow.

This is a bit of a tricky UAT request because we need to make a whole release. Once an icon is decided on, an ENUM property will likely be unnecessary and we can clean this up. If we decide to go with one of these new options, it may also be worth considering to source all icons from Carbon rather than mixing libraries? 

| `resetControlImage` value  | display   |
|---|---|
| `unicode` (default) | Screenshot from 2022-10-14 10-11-37 |
| `reset`  |  Screenshot from 2022-10-14 10-09-53 |
|  `trash` | Screenshot from 2022-10-14 10-10-16 |
| `erase` | Screenshot from 2022-10-14 10-10-30 |

### DIFF
--- a/src/components/my-map/controls.ts
+++ b/src/components/my-map/controls.ts
@@ -1,6 +1,9 @@
-import "ol/ol.css";
 import { Control, ScaleLine } from "ol/control";
+import "ol/ol.css";
+import eraseIcon from "./icons/erase.svg";
 import northArrowIcon from "./icons/north-arrow-n.svg";
+import resetIcon from "./icons/reset.svg";
+import trashCanIcon from "./icons/trash-can.svg";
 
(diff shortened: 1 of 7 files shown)


jessicamcinchak:
[What should we change draw outline to? icon from ↩️ to ?](https://trello.com/c/24g66bw4/2112-what-should-we-change-draw-outline-to-icon-from-%E2%86%A9%EF%B8%8F-to)

DafyddLlyr (approved):
Looks good!

Agree it's a tricky / slow one to test. I think a single icon library is probably best from realistically to try to maintain a standard look and feel.

Let's see what the feedback is!

PR #230, opened 2022-12-07
chore: cleanup styling of default scale line

jessicamcinchak:
Override the default openlayers translucent blue with a cleaner style that better fits our Gov UK-esque styles
Screenshot from 2022-12-07 13-00-18


jessicamcinchak:
[Add new north point and scale bar styles to Map component](https://trello.com/c/sCCtRzii/2226-add-new-north-point-and-scale-bar-styles-to-map-component)

PR #238, opened 2022-12-14
fix: Display static geojson display in docs

DafyddLlyr:
image


jessicamcinchak (approved):
:tada: thanks for fixing this long-lingering bug!! (a quick re-run of the vitest action will hopefully clear up the timeout issue on your first run, I've seen that one before too)

PR #239, opened 2022-12-14
fix: Resolve shadow class fields warning

DafyddLlyr:
A warning is being thrown in the dev sandbox console, pointing to the following docs -  https://lit.dev/docs/components/properties/\#avoiding-issues-with-class-fields

Possibly something picked up after the most recent updates?

image


jessicamcinchak (approved):
Thank you! :+1:

PR #241, opened 2022-12-14
feat: Hide OS API key by allowing proxy endpoint

DafyddLlyr:
# What does this PR do?
- Support the use of a proxy via the `osProxyEndpoint` property
- This means that we do not need to expose the OS API key when making requests

## How is this achieved?
Before making any requests to the OS, we get the URL from `ordnanceSurvey.getServiceURL()`. 

If the user supplied an OS API key, we'll query the OS API directly and append their key. This is the exact same functionality as currently exists.

If the user supplies an `osProxyEndpoint` value, we'll construct the correct service URL, without an API key. It is the end user's responsibility to host and maintain a suitable service to proxy requests (with the correct authentication) to the OS API.  Ensuring proper access to the proxy endpoint can be done through CORS/CORP on the proxy. The documentation in this repository outlines how this can be achieved.

## Testing
In order to test this, you can run the dev server locally and add the `osProxyEndpoint="https://api.planx.dev/proxy/ordnance-survey"` property to the `my-map` or `address-autocomplete` components. This will also work on the ["Deploy Preview" link](https://deploy-preview-241--oslmap.netlify.app/) generated by the PR.

A number of unit test and service / integration tests have also been added to cover the new functionality, but there's plenty more that could be added to cover all scenarios. I think I've captured the main journeys here but please let me know if you think we need more coverage anywhere.

## Documentation
Pitsby doesn't allow markup or any rich text (even line breaks) in it's descriptions, so linking back to a richer markdown page seemed helpful here.

### DIFF
--- a/docs/how-to-use-a-proxy.md
+++ b/docs/how-to-use-a-proxy.md
@@ -1,0 +1,97 @@
+# How to: Use a MyMap & AddressAutocomplete with a proxy
+
+## Context
+Both `MyMap` and `AddressAutocomplete` can call the Ordnance Survey API directly, or via a proxy. 
+
+Calling the API directly may be suitable for internal use, where exposure of API keys is not a concern, whilst calling a proxy may be more suitable for public use.
+
+A proxy endpoint can be supplied via the `osProxyEndpoint` property on these components.
+
+Proxies are required to complete the following actions in order to work successfully - 
+
+- Append a valid OS API key as a search parameter to incoming requests
+- Modify outgoing response with suitable CORS / CORP headers to allow the originating site access to the returned assets
+
+## Diagram
+```mermaid
+sequenceDiagram
+    autonumber
+    participant WC as Web Component
+    participant P as Proxy
+    participant OS as Ordnance Survey API
+
+    WC ->>+ P: Request
+    P -->> P: Validate Request
+    P ->>+ OS: Request + API key
+    OS ->>- P: Response
+    P ->>- WC: Response + CORP/CORS headers
+```
+
+## Examples
+Please see the sample code below for how a proxy could be implemented - 
+
+### Express
+Below is an annotated example of a simple proxy using [Express](https://github.com/expressjs/express) & [http-proxy-middleware](https://github.com/chimurai/http-proxy-middleware).
+
+**index.js**
+```js
+import express from "express";
+import { useOrdnanceSurveyProxy } from "proxy";
+
+const app = express()
+const port = 3000
+
+app.use('/proxy/ordnance-survey', useOrdnanceSurveyProxy)
+
+app.listen(port)
+```
+
+**proxy.js**
+```js
+import { createProxyMiddleware } from "http-proxy-middleware";
+
+const OS_DOMAIN = "https://api.os.uk";
+
+export const useOrdnanceSurveyProxy = async (req, res, next) => {
+  if (!isValid(req)) return next({
+    status: 401,
+    message: "Unauthorised"
+  })
+
+  return createProxyMiddleware({
+    target: OS_DOMAIN,
+    changeOrigin: true,
+    onProxyRes: (proxyRes) => setCORPHeaders(proxyRes),
+    pathRewrite: (fullPath, req) => appendAPIKey(fullPath, req)
+    onError: (_err, _req, res) => {
+      res.json({
+        status: 500,
+        message: "Something went wrong",
+      });
+    },
+  })(req, res, next);
+};
+
+const isValid = (req) => {
+  // Your validation logic here, for example checking req.header.referer against an allowlist of domains
+}
+
+// Ensure that returned tiles can be embedded cross-site
+// May not be required if "same-site" policy works for your setup
+const setCORPHeaders = (proxyRes: IncomingMessage): void => {
+  proxyRes.headers["Cross-Origin-Resource-Policy"] = "cross-origin"
+}
+
+export const appendAPIKey = (fullPath, req) => {
+  const [path, params] = fullPath.split("?");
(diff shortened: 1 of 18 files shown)


index.html
```diff
@@ -43,7 +43,7 @@
     <div style="display:flex;flex-direction:column;">
       <h1 style="color:red;font-family:Inter,Helvetica,sans-serif;font-size:16px;">*** This is a testing sandbox - these components are unaware of each other! ***</h1>
       <div style="margin-bottom:1em">
-        <my-map zoom="20" maxZoom="23" drawMode drawPointer="dot" id="example-map" showScale />
+        <my-map zoom="20" maxZoom="23" drawMode drawPointer="dot" id="example-map" showScale disableVectorTiles osProxyEndpoint="http://api.1334.planx.pizza/proxy/ordnance-survey" />
```
jessicamcinchak, line 46: I think `osProxyEndpoint` can point to planx production or staging, rather than a pizza link, here now?

docs/how-to-use-a-proxy.md
```diff
@@ -0,0 +1,97 @@
...
+  const updatedParams = new URLSearchParams(params);
+  updatedParams.set("key", process.env.ORDNANCE_SURVEY_API_KEY);
+  // Remove our API baseUrl (/proxy/ordnance-survey)
+  const updatedPath = path.replace(req.baseUrl, "");
+  // Construct and return rewritten path
+  const resultPath = [updatedPath, updatedParams.toString()].join("?");
+  return resultPath;
+};
+```
+> A working and more fleshed out example (in TypeScript) can be seen [here in the PlanX API](https://github.com/theopensystemslab/planx-new/blob/production/api.planx.uk/proxy/ordnanceSurvey.ts).
```
jessicamcinchak, line 97: Very thorough docs, thank you!

src/components/address-autocomplete/main.test.ts
```diff
@@ -80,3 +80,37 @@ describe("AddressAutocomplete on initial render with empty postcode", async () =
...
+
+  it("calls OS API when 'osPlacesApiKey' provided", async () => {
+    document.body.innerHTML = `<address-autocomplete id="autocomplete-vitest" postcode="SE5 0HU" osPlacesApiKey=${
+      import.meta.env.VITE_APP_OS_PLACES_API_KEY
+    } />`;
+    await window.happyDOM.whenAsyncComplete();
+
+    expect(fetchSpy).toHaveBeenCalledWith(
+      expect.stringContaining(
+        "https://api.os.uk/search/places/v1/postcode?postcode=SE5+0HU"
```
jessicamcinchak, line 112: can we tack on `"&key="` to this string so it's super clear that we expect the API key directly appended to the request URL in this case? 

and can we use a dummy api key value like "secret" in this mock rather than our env var because that might be publicly logged while the vitest Github Action is running ?? log looks like only a summary now, but I'd guess it's verbose for the few seconds while it's active?

src/components/my-map/os-layers.ts
```diff
@@ -5,46 +5,89 @@ import VectorTileLayer from "ol/layer/VectorTile";
...
+  const isUsingOS = Boolean(apiKey || proxyEndpoint);
+  // Fallback to OSM if not using OS services
+  const basemap = isUsingOS
+    ? makeOSRasterBaseMap(apiKey, proxyEndpoint, copyright)
+    : makeOSMRasterBasemap();
+  basemap.set("name", "rasterBaseMap");
+  return basemap;
+}
+
+function makeOSRasterBaseMap(
```
jessicamcinchak, line 24: these two function names are pretty confusing for me - is this second one perhaps better described as `getTileLayer` or something?

src/lib/ordnanceSurvey.test.ts
```diff
@@ -0,0 +1,84 @@
...
+      params: { srs: "3857" },
+    });
+
+    expect(result).toBeDefined();
+    const { origin, pathname, searchParams } = new URL(result!);
+    expect(origin).toEqual("https://www.my-site.com");
+    expect(decodeURIComponent(pathname)).toEqual(
+      "/api/proxy/os/maps/vector/v1/vts/resources/styles"
+    );
+    expect(searchParams.get("srs")).toEqual("3857");
```
jessicamcinchak, line 57: similar to a comment above, would it be a bit stronger to additionally include a check that `searchParams` does _not_ contain "key" param at all in this proxy case.

`expect(searchParams.get("key")).toBeUndefined;` maybe?

src/lib/ordnanceSurvey.ts
```diff
@@ -0,0 +1,83 @@
...
+  | "vectorTile"
+  | "vectorTileStyle"
+  | "places"
+  | "features";
+interface ServiceOptions {
+  service: OSServices;
+  apiKey: string;
+  proxyEndpoint: string;
+  params?: Record<string, string>;
+}
```
jessicamcinchak, line 13: nit: can you add an empty line between these definitions please for just a bit easier readability :sweat_smile:

otherwise this lib file looks great !

jessicamcinchak (commented):
Looking really solid :tada: I want to come back with fresher eyes at least once more, but here's initial comments/questions. Appreciate the docs, examples and tests here - makes it much easier to review!

index.html
DafyddLlyr: Yep - missed in the tidy up, fixed!

src/components/my-map/os-layers.ts
DafyddLlyr: Yeah this is a very fair comment - names updated.

jessicamcinchak (approved):

PR #261, opened 2023-02-06
fix: correctly set geojson writeFeaturesObject options for coordinates projected in EPSG:27700

jessicamcinchak:
fixes bug spotted over here: https://github.com/theopensystemslab/planx-new/pull/1377#discussion_r1091647359

geojson change events now logging out "EPSG:27700" coordinates as proper 6-digit positive integers, "EPSG:3857" coordinates were always correct - and remain so!

### DIFF
--- a/src/components/my-map/utils.ts
+++ b/src/components/my-map/utils.ts
@@ -77,7 +77,6 @@ export function makeGeoJSON(
       ? {
           dataProjection: projection,
           featureProjection: "EPSG:3857",
-          decimals: 0,
         }
       : { featureProjection: projection };
 
--- a/src/components/my-map/utils.ts
+++ b/src/components/my-map/utils.ts
@@ -71,10 +71,18 @@ export function makeGeoJSON(
   source: VectorSource<Geometry> | Feature<Geometry>[],
   projection: ProjectionEnum
 ): GeoJSONObject {
+  // ref https://openlayers.org/en/latest/apidoc/module-ol_format_GeoJSON-GeoJSON.html#writeFeaturesObject
+  const options =
+    projection === "EPSG:27700"
+      ? {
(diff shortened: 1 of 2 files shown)


src/components/my-map/utils.ts
```diff
@@ -71,10 +71,18 @@ export function makeGeoJSON(
   source: VectorSource<Geometry> | Feature<Geometry>[],
   projection: ProjectionEnum
 ): GeoJSONObject {
+  // ref https://openlayers.org/en/latest/apidoc/module-ol_format_GeoJSON-GeoJSON.html#writeFeaturesObject
+  const options =
+    projection === "EPSG:27700"
+      ? {
+          dataProjection: projection,
+          featureProjection: "EPSG:3857",
+          decimals: 0,
```
DafyddLlyr, line 80: So this is a tricky one for me here.

`decimals: 0` will (_I think_) give us a less accurate/meaningful GeoJSON, because BNG Eastings / Northings to 8 figures are accurate to 10m (https://digimap.edina.ac.uk/help/our-maps-and-data/bng/). 

What I think we want to do is still output at the drawn accuracy (as well as OpenLayers can give us), but then just display in PlanX a rounded, consumable version?

DafyddLlyr (changes requested):
One small question / comment!

src/components/my-map/utils.ts
jessicamcinchak: this is fair! while doing this, I did confirm that OS returns non-decimal BNG coordinates via OS Places, but I'm happy to keep decimal accuracy for ourselves here and just display a rounded version later for max accuracy. updated now!

DafyddLlyr (approved):

PR #263, opened 2023-02-15
feat: Printing

DafyddLlyr:
Apologies that this one has taken so long to see any daylight 🌤️ I kept coming back and tweaking... it's pretty hacky in places.

I'm using `ol-ext/printDialog` ([docs](https://viglino.github.io/ol-ext/doc/doc-pages/ol.control.PrintDialog.html)) to allow the map to be printed to PDF/PNG/JPG at set scales. 

The documentation is pretty poor for `ol-ext` so there's been a lot of figuring out and reverse engineering.

This meets the requirements of the following ticket (and then some) -  https://trello.com/c/jVS79IdF/2230-can-automatically-generated-location-plan-click-to-standard-scales

image

image

**To do...**
 - [x] Pitsby docs
 - [ ] Unsquash print compass
 - [ ] Handle imperfect scales
 - [x] Style: Handle button position if "Reset" button not present


index.html
```diff
@@ -39,11 +39,11 @@
       }
     </style> -->
   </head>
-  <body>
+  <body style="font-family:Inter,Helvetica,sans-serif;">
```
DafyddLlyr, line 42: This was moved to test that font style is inherited in the `PrintDialog`

src/components/my-map/controls.ts
```diff
@@ -48,3 +54,115 @@ export function resetControl(listener: any, icon: string) {
 
   return new Control({ element: element });
 }
+
+PrintDialog.prototype.scales = {
```
DafyddLlyr, line 58: There's no way of setting this within the class - https://github.com/Viglino/ol-ext/issues/766#issuecomment-1108267184

src/components/my-map/controls.ts
```diff
@@ -48,3 +54,115 @@ export function resetControl(listener: any, icon: string) {
...
+      .getArray()
+      .filter(
+        (control) => control instanceof CanvasScaleLine
+      )[0] as CanvasScaleLine;
+    const canvasScaleLine = document.querySelector(
+      ".ol-scale-line-inner"
+    ) as HTMLDivElement;
+    const scale = Math.round(scaleLineControl.getScaleForResolution());
+    if (canvasScaleLine) {
+      canvasScaleLine.innerHTML = `1 : ${scale}`;
```
DafyddLlyr, line 157: Frustratingly, this renders a scale which doesn't _quite_ match the dropdown options.

I've not done too deep a dive, but I'm wondering if it's a rounding issue in ol-ext (seems pretty unlikely?), or something to do with how the `CanvasScaleLine` is rendered (pixel width must match `ScaleLine`).

We could fix this with rounding, but I'm not sure how kosher that would be...!

src/components/my-map/controls.ts
```diff
@@ -48,3 +54,115 @@ export function resetControl(listener: any, icon: string) {
...
+      this.getMap()?.on("postrender", () => this.syncScaleText());
+    });
+  }
+
+  /**
+   * Display scale (1:XXX) on CanvasScaleLine control
+   * This can not natively be done with this element, but it has all the inherited
+   * functions from OL ScaleLine to allow us to set this
+   */
+  private syncScaleText() {
```
DafyddLlyr, line 145: I spent a good while trying to make this a method on the `CanvasScaleLine` class (or rather a child class) as really that's where it belongs and you can access the `this.element` without query selectors etc.

The problem I hit was getting the correct map context from within than class, even when explicitly passing `this.map` into the constructor.

src/components/my-map/os-layers.ts
```diff
@@ -35,6 +35,7 @@ function makeOSRasterBaseMap(
     source: new XYZ({
       url: tileServiceURL,
       attributions: [copyright],
+      crossOrigin: "anonymous",
```
DafyddLlyr, line 38: This is a requirement for the canvas to access tiles - https://github.com/Viglino/ol-ext/issues/673#issuecomment-937674728

jessicamcinchak (approved):
I'm sorry it's taken me so long to properly get to this one - I've looked at it many times and failed to hit "comment"! 

Overall I'm happy with this solution and it functions perfectly for me, it's actually a lot less code than I expected it would be too. Solid work for a big-feeling feature :printer: :tada: 

A couple general comments / questions (happy to chat about any of these outloud, please don't feel any pressure to type responses): 
- You added "Pitsby examples" in the todo's, but I wonder if there's actually a case to intentionally omit an expample of the print button from docs because otherwise anyone could theoretically print OS basemaps via the "Sandbox" using our proxy endpoint? Alastair comment ed about this legal difference in web maps versus print a bit ago if you recall?
- I'm curious about reasons you chose `ol-ext`, rather than say using the `html2canvas` & `jsPDF` approach in the openlayers examples https://openlayers.org/en/latest/examples/print-to-scale.html (obvious downside being would need to roll our own dialog inputs I guess)? With the funny scale issue going on, I'm curious about effort involved to sense check against a different library and see if clears it up - or if it's just a quirk of printing we're going to live with (I'm also totally OK with this if councils are and will be curious to see if it gets noticed in UAT)

src/components/my-map/index.ts
```diff
@@ -260,12 +270,29 @@ export class MyMap extends LitElement {
...
-      map.addControl(northArrowControl());
+    if (this.showPrint) {
+      const printControl = new PrintControl();
+      map.addControl(printControl);
+
+      // Toggle scaleControl when printControl is open
+      // Instead, display CanvasScaleLine which can be printed
+      if (this.showScale) {
+        // @ts-ignore
+        printControl.on("show", () => map.removeControl(scale));
```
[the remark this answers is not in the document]
DafyddLlyr: This is a fair comment 👍 Found a way of resolving this by passing a reference to the main map into the control.

src/components/my-map/controls.ts
```diff
@@ -48,3 +54,115 @@ export function resetControl(listener: any, icon: string) {
...
+    const scaleLineControl = this.getMap()
+      ?.getControls()
+      .getArray()
+      .filter(
+        (control) => control instanceof CanvasScaleLine
+      )[0] as CanvasScaleLine;
+    const canvasScaleLine = document.querySelector(
+      ".ol-scale-line-inner"
+    ) as HTMLDivElement;
+    const scale = Math.round(scaleLineControl.getScaleForResolution());
```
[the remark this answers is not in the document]
DafyddLlyr: Screenshot 2023-02-24 at 12 55 32

Without rounding we get number which is not particularly meaningful, and by convention map scales are generally displayed as integers.

src/components/my-map/controls.ts
DafyddLlyr: I'll raise an issue for this in the repo, and will move forward with testing for now.

DafyddLlyr:
@jessicamcinchak To summarise our conversation on this - 
 - No docs (for now) is the way to go as there's just a single specific use case
 - `ol-ext` gives us a nice dialog and preview map, and allows for a decent amount of customisation without too much overhead. I had thought that canvas scale bars would be an issue, but this isn't actually the case in the linked example.

PR #266, opened 2023-02-24
v0.8.0

DafyddLlyr:


jessicamcinchak (changes requested):
since we're still pre-v1.0.0, minor version is only implemented upon a breaking change & patch version will be incremented for features! so I think this is actually v0.7.2

DafyddLlyr:
Closed in favour of https://github.com/theopensystemslab/map/pull/267

PR #360, opened 2023-08-08
build: update node and pnpm versions & drop package.json `engines`

jessicamcinchak:
It's nice to be able to use the same versions as `planx-new` and not have to switch between dev environments


package.json
```diff
@@ -60,8 +60,8 @@
     "wait-for-expect": "^3.0.2"
   },
   "engines": {
-    "node": "^16",
-    "pnpm": "^7.8.0"
+    "node": "^18",
+    "pnpm": "^8.6.0"
```
DafyddLlyr, line 64: We've actually dropped `engines` from other repos - we can probably do that here also.

See https://github.com/theopensystemslab/planx-new/pull/1918 for explanation 👍

jessicamcinchak: Thanks, I did remember this but was hesitant because Netlify logs seemed to be referencing `engines` in the build setup - but all CI succeeding now that it's dropped :checkered_flag:

DafyddLlyr (approved):

PR #363, opened 2023-08-13
feat: clip viewport based on geojson extent

jessicamcinchak:
Introduces an optional prop `clipGeojsonData` that can be set to disable a user panning/navigating outside of their own council. 

I originally was trying to apply a clipping/mask layer but I think that requires more style tinkering than its worth in our case, and this simpler option of applying an extent to the viewport should satisfy what we need ! 

See these docs / examples to compare:
- https://openlayers.org/en/latest/examples/extent-constrained.html
- https://openlayers.org/en/latest/examples/layer-clipping-vector.html

Next steps:
- Review/merge this PR
- Queue up new map release https://github.com/theopensystemslab/map/pull/364
- Open a planx-new branch with the upgrade
  - Set the `clipGeojsonData` based on the existing `teams.setting.boundary` geojson if it exists
  - Switch off static mode for document maps and generate a sample zip (`planx-core`)
  - Open pizza for user-testing (prompts to check addresses at border especially, does extent and current zoom still allow bordered addresse to appear "centered" etc? Are any valid addresses "unreachable" because of the extent?)
  
Ideally this can sit in user-testing while Jess is out this next week, then get finalised/merged next week. At which point we could also decide to revisit if our current storage of `teams.settings.boundary` is fine here or if we want to tweak that too in the background.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,19 +1,18 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="UTF-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>OS Web Components Sandbox</title>
-    <script type="module" src="./src/index.ts"></script>
-    <!-- OS vector tile source specifies fonts in .pbf format, which OpenLayers can't load, so make them available directly  -->
-    <link rel="preconnect" href="https://fonts.googleapis.com" />
-    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
-    <link
-      href="https://fonts.googleapis.com/css2?family=Source+Sans+Pro:wght@300;400;600&display=swap"
-      rel="stylesheet"
-    />
-    <!-- Examples of available style options for postcode-search & address-autocomplete -->
-    <!-- <style>
+
+<head>
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <title>OS Web Components Sandbox</title>
+  <script type="module" src="./src/index.ts"></script>
+  <!-- OS vector tile source specifies fonts in .pbf format, which OpenLayers can't load, so make them available directly  -->
+  <link rel="preconnect" href="https://fonts.googleapis.com" />
+  <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+  <link href="https://fonts.googleapis.com/css2?family=Source+Sans+Pro:wght@300;400;600&display=swap"
+    rel="stylesheet" />
+  <!-- Examples of available style options for postcode-search & address-autocomplete -->
+  <!-- <style>
       address-autocomplete {
         --autocomplete__label__font-size: 25px;
         --autocomplete__input__padding: 6px 40px 7px 12px;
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -59,58 +61,89 @@ <h1 style="color:red;font-size:16px;">*** This is a testing sandbox - these comp
...
+              51.50705957656797
+            ],
+            [
+              -0.128307852848053,
+              51.50748361634746
+            ]
+          ]
+        ]
+      },
+    };
```
jessicamcinchak, line 99: lots of linter changes here, but this is the notable change to demo the clipped viewport !

src/components/my-map/index.ts
```diff
@@ -223,20 +229,27 @@ export class MyMap extends LitElement {
...
-          "EPSG:3857"
-        ),
+        extent: clipExtent
+          ? clipExtent
+          : transformExtent(
+              // UK Boundary
+              [-10.76418, 49.528423, 1.9134116, 61.331151],
+              "EPSG:4326",
+              "EPSG:3857",
+            ),
```
jessicamcinchak, line 252: again, lots of linter changes here, but this is the juicy bit !

src/components/my-map/index.ts
```diff
@@ -189,6 +189,14 @@ export class MyMap extends LitElement {
   @property({ type: Boolean })
   showPrint = false;
 
+  @property({ type: Object })
+  clipGeojsonData = {
+    type: "Feature",
+    geometry: {
+      coordinates: [],
+    },
+  };
```
jessicamcinchak, line 198: Lit default prop values for optional props can feel a bit awkward, espcially for complex types like geojson object. Might be worth revisiting this in a future PR to touch all the geojson props at once :bookmark:

jessicamcinchak:
Netlify deploys are passing on retry fine but not updating Github CI here - not positive what's going on there yet https://app.netlify.com/sites/oslmap/deploys/64d897e517608f7a1ff157ac

jessicamcinchak:
[Switch zooming the red edge on the HTML / map extract back on](https://trello.com/c/uqh888JE/2528-switch-zooming-the-red-edge-on-the-html-map-extract-back-on)

src/components/my-map/index.ts
```diff
@@ -189,6 +189,14 @@ export class MyMap extends LitElement {
   @property({ type: Boolean })
   showPrint = false;
 
+  @property({ type: Object })
```
DafyddLlyr, line 192: A simpler API might be to accept an [extent](https://openlayers.org/en/latest/apidoc/module-ol_extent.html#~Extent) e.g. `[minx, miny, maxx, maxy]`. We could then default to the UK extent here also.

It does put the onus on the consumer of this webcomponent to generate one however which may not suit BOPS as well as it suits us?

src/components/my-map/index.ts
```diff
@@ -223,20 +231,29 @@ export class MyMap extends LitElement {
...
-      "EPSG:3857"
+      "EPSG:3857",
     );
 
+    const clipFeature =
+      this.clipGeojsonData.geometry?.coordinates?.length > 0 &&
+      new GeoJSON().readFeature(this.clipGeojsonData, {
+        featureProjection: "EPSG:3857",
+      });
+    const clipExtent = clipFeature && clipFeature.getGeometry()?.getExtent();
```
DafyddLlyr, line 242: I suspect that we'll need to buffer this somewhat to get the functionality we want - the question is who's responsibility is it?

If a consumer adds a "clip" geojson but the map then shows more than they're clip that's quite odd / bad behaviour. We could add a buffer / scale value as an additional prop, or just leave as is and leave the responsibility to consumers.

I think I'd probably also go with not scaling / buffering here 👍 This might be something to specifically mention in docs / to BOPS though ("you probably want to buffer or scale the geometry to give a better user experience for those on the edges of areas").

DafyddLlyr (approved):
Pulled the branch locally, installed dependencies and then updated lockfile and all worked as expected with CI.

PR is great - two small comments to discuss but I suspect straight GeoJSON is the right fit here.

Really appreciate the helpful comments and having the list of follow up tasks laid out so clearly - super helpful. I'll start picking them up so we can get this on a Pizza this sprint.

PR #379, opened 2023-08-29
Allow drawing colour and fill to be customisable

benbaumann95:
### Description of change
- In [Bops](https://github.com/unboxed/bops), we want to be able to draw a polygon on the map to visualise a consultation area where we return the addresses in that drawn area. However, this is currently hardcoded in red and to distinguish between the red line boundary and consultation area we preferably want this in a different colour that can be customisable. This PR adds this ability.
- The drawingColor and drawingColorFill defaults to what we usually expect for the red line boundary

Trello ticket:
https://trello.com/c/IQsb3WNF/1908-improve-content-and-layout-for-querying-neighbour-addresses-by-polygon-search

Screenshot 2023-08-30 at 14 03 56

### DIFF
--- a/src/components/my-map/drawing.ts
+++ b/src/components/my-map/drawing.ts
@@ -11,78 +11,91 @@ export type DrawTypeEnum = "Polygon" | "Point"; // ref https://openlayers.org/en
 export type DrawPointerEnum = "crosshair" | "dot";
 
 // drawPointer styles
-const crosshair = new RegularShape({
-  stroke: new Stroke({
-    color: "red",
-    width: 2,
-  }),
-  points: 4, // crosshair aka star
-  radius1: 15, // outer radius
-  radius2: 1, // inner radius
-});
-
-const dot = new CircleStyle({
-  radius: 6,
-  fill: new Fill({
-    color: "#ff0000",
-  }),
-});
-
-// feature style: red-line site boundary
-const redLineBase = {
-  color: "#ff0000",
-  width: 3,
-};
-
-const redLineStroke = new Stroke(redLineBase);
-
-const redDashedStroke = new Stroke({
-  ...redLineBase,
-  lineDash: [2, 8],
-});
-
-const redLineFill = new Fill({
-  color: "rgba(255, 0, 0, 0.1)",
-});
-
-const polygonVertices = new Style({
-  image: new RegularShape({
-    fill: new Fill({
-      color: "#fff",
-    }),
+function crosshair(drawColor: string) {
+  return new RegularShape({
     stroke: new Stroke({
-      color: "#ff0000",
+      color: drawColor,
       width: 2,
     }),
-    points: 4, // squares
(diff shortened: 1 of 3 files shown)


src/components/my-map/drawing.ts
```diff
@@ -70,22 +53,46 @@ const polygonVertices = new Style({
...
+    new Style({
+      fill: new Fill({
+        color: drawingColorFill,
+      }),
+      stroke: new Stroke({
+        color: drawingColor,
+        width: 3,
+      }),
+    }),
+    polygonVertices,
```
jessicamcinchak, line 70: nit: can `polygonVertices` also be refactored as a small function that takes the `drawingColor` argument ? They're still hardcoded to red!

jessicamcinchak (commented):
Thanks for these changes, a very useful feature! Two minor comments, but otherwise looks great :crayon:

src/components/my-map/docs/my-map-features.doc.js
```diff
@@ -13,6 +13,16 @@ module.exports = {
...
+    {
+      name: "drawingColor",
+      type: "String",
+      values: "",
+    },
+    {
+      name: "drawingFillColor",
+      type: "String",
+      values: "",
+    },
```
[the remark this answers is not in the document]
benbaumann95: Oh yeah woopsie

src/components/my-map/drawing.ts
```diff
@@ -28,64 +28,73 @@ const dot = new CircleStyle({
...
+  drawingFillColor: string,
+) {
   return new Style({
-    stroke: redDashedStroke,
-    fill: redLineFill,
+    stroke: configureDashedStroke(drawingColor),
+    fill: new Fill({
+      color: drawingFillColor,
+    }),
     image: pointerStyle === "crosshair" ? crosshair : dot,
```
jessicamcinchak, line 86: some lingering hard-coded "red" sorry: can crosshair & dot also reflect the configured color prop?

src/components/my-map/index.ts
```diff
@@ -108,6 +108,12 @@ export class MyMap extends LitElement {
   @property({ type: Boolean })
   clickFeatures = false;
 
+  @property({ type: String })
+  drawingColor = "#ff0000";
+
+  @property({ type: String })
+  drawingFillColor = "rgba(255, 0, 0, 0.1)";
```
jessicamcinchak, line 115: nit: apologies I didn't catch this first time around, but can these prop name prefixes please be `draw` instead of `drawing` for consistency with existing draw-related props?

jessicamcinchak (commented):
Sorry two more minor tweaks that I just caught on final review! 

Once adjusted, I'm happy to approve and merge this and make the release (limited permissions to that step because OSL NPM account) and share new version number with you for use :tada: Thanks again for this!

jessicamcinchak (approved):

PR #381, opened 2023-08-31
Set color of polygon with color set in the geojson feature properties

benbaumann95:
- Default to geojsonColor if the properties color object does not exist
- Can be added to feature i.e. `"properties": {"color": "#e670fc"}`

https://trello.com/c/IQsb3WNF/1908-improve-content-and-layout-for-querying-neighbour-addresses-by-polygon-search

Screenshot 2023-08-31 at 12 45 59

### DIFF
--- a/src/components/my-map/index.ts
+++ b/src/components/my-map/index.ts
@@ -374,17 +374,22 @@ export class MyMap extends LitElement {
 
     const geojsonLayer = new VectorLayer({
       source: geojsonSource,
-      style: new Style({
-        stroke: new Stroke({
-          color: this.geojsonColor,
-          width: 3,
-        }),
-        fill: new Fill({
-          color: this.geojsonFill
-            ? hexToRgba(this.geojsonColor, 0.2)
-            : hexToRgba(this.geojsonColor, 0),
-        }),
-      }),
+      style: function (this: MyMap, feature: any) {
+        // Retrieve color from feature properties
+        let featureColor = feature.get("color") || this.geojsonColor; // Use the geojsonColor if no color property exists
+
+        return new Style({
+          stroke: new Stroke({
+            color: featureColor,
+            width: 3,
+          }),
+          fill: new Fill({
+            color: this.geojsonFill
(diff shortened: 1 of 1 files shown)


src/components/my-map/index.ts
```diff
@@ -374,17 +374,22 @@ export class MyMap extends LitElement {
...
-          color: this.geojsonColor,
-          width: 3,
-        }),
-        fill: new Fill({
-          color: this.geojsonFill
-            ? hexToRgba(this.geojsonColor, 0.2)
-            : hexToRgba(this.geojsonColor, 0),
-        }),
-      }),
+      style: function (this: MyMap, feature: any) {
```
benbaumann95, line 377: Was struggling to add the specific type for `feature`.
Tried adding `feature: Feature` and `feature: typeof Feature` with no luck even though this is an instance of Feature

jessicamcinchak: No worries here, OpenLayers can be finiky about this - it _might_ be: 
```ts
import { Feature } from "ol/index";

style: function (this: MyMap, feature: Feature<Geometry>) {...}
```
But if that still complains, feel free to leave this as `any` and I can re-visit later!

src/components/my-map/index.ts
```diff
@@ -374,17 +374,22 @@ export class MyMap extends LitElement {
...
+        // Retrieve color from feature properties
+        let featureColor = feature.get("color") || this.geojsonColor; // Use the geojsonColor if no color property exists
```
jessicamcinchak, line 379: :raised_hands: smart & simple !

jessicamcinchak (approved):
Thanks again for this ! Tested and works perfectly :art: 

Left a small type suggestion, but no pressure to spend time on that if not straightforward. Happy for this to be "Squashed & merged" anytime and I'll get the new release through before end of day for you!

src/components/my-map/index.ts
benbaumann95: Yeah still complaining 😕 .
Let me know if you do figure this one out, i'm intrigued now !

PR #429, opened 2024-01-24
feat: Show neighbouring properties as points on map

cdccollins:
Creates a property showNeighbourMark which when switched on creates a marker for each point passed in on the geojsonData property.

This will be useful in BOPS as it can be difficult to know when drawing a polygon whether you have selected the neighbouring properties. BOPS will pass the coordinates of the neighbours via the geojsonData prop.

### DIFF
--- a/src/components/my-map/index.ts
+++ b/src/components/my-map/index.ts
@@ -50,6 +50,7 @@ import {
   hexToRgba,
   makeGeoJSON,
 } from "./utils";
+import { GeoJSONFeature } from "ol/format/GeoJSON";
 
 type MarkerImageEnum = "circle" | "pin";
 type ResetControlImageEnum = "unicode" | "trash";
@@ -130,7 +131,10 @@ export class MyMap extends LitElement {
   featureBuffer = 40;
 
   @property({ type: Boolean })
-  showMarker = false;
+  showGeojsonDataMarkers = false;
+
+  @property({ type: Boolean })
+  showCentreMarker = false;
 
   @property({ type: String })
   markerImage: MarkerImageEnum = "circle";
@@ -587,14 +591,9 @@ export class MyMap extends LitElement {
       }
     };
 
-    // show a marker at a point
-    if (this.showMarker) {
+    const showNewMarker = (lon: number, lat: number) => {
       const markerPoint = new Point(
-        transform(
-          [this.markerLongitude, this.markerLatitude],
-          projection,
-          "EPSG:3857",
-        ),
+        transform([lon, lat], projection, "EPSG:3857"),
       );
       const markerLayer = new VectorLayer({
         source: new VectorSource({
@@ -604,6 +603,20 @@ export class MyMap extends LitElement {
       });
 
       map.addLayer(markerLayer);
+    };
+
+    // show a marker at a point
+    if (this.showCentreMarker) {
+      showNewMarker(this.markerLongitude, this.markerLatitude);
+    }
+
+    if (this.showGeojsonDataMarkers) {
+      this.geojsonData.features.forEach((feature: GeoJSONFeature) => {
+        showNewMarker(
+          feature.geometry.coordinates[0],
+          feature.geometry.coordinates[1],
+        );
+      });
     }
 
     // XXX: force re-render for safari due to it thinking map is 0 height on load


src/components/my-map/index.ts
```diff
@@ -129,6 +129,9 @@ export class MyMap extends LitElement {
   @property({ type: Number })
   featureBuffer = 40;
 
+  @property({ type: Boolean })
+  showNeighbourMarkers = false;
```
jessicamcinchak, line 133: nit: since this prop reads from `geojsonData`, maybe naming as `showGeojsonDataMarkers` would offer more clarity on when to use it? The component itself doesn't necessarily need to know about the concept/use-case of "neighbours" I think / it works for any collection of dispersed points! 

(`showMarker` &rarr; `showCentreMarker` might be nice clarification for existing prop too now that we're doubling down on this feature - but not sure it's worth a breaking change! Your call, easy enough for us to adjust our map instances on next upgrade if updated :slightly_smiling_face:)

src/components/my-map/index.ts
```diff
@@ -604,6 +602,20 @@ export class MyMap extends LitElement {
...
       map.addLayer(markerLayer);
+    };
+
+    // show a marker at a point
+    if (this.showMarker) {
+      showNewMarker(this.markerLongitude, this.markerLatitude);
+    }
+
+    if (this.showNeighbourMarkers) {
+      this.geojsonData.features.forEach((feature: any) => {
```
jessicamcinchak, line 613: nit: rather than `any` this should be able to use type `GeoJSONFeature` from `"ol/format/GeoJSON"`

jessicamcinchak (commented):
Thanks for this new feature !! :tada: 

Only one small question while reviewing: when I pass `geojsonData` and turn on `showNeighbourMarkers` it's not actually displaying the default black cirlces or any other image (it does correctly re-center the map over where the points _should_ be)? I can still see the defualt marker image fine when just displaying the singular `showMarker`. 

Is it the same for you / is there something different about my example data than what you plan to pass in? 
```
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "coordinates": [
          -0.13708768238623747,
          51.51141463106259
        ],
        "type": "Point"
      }
    },
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "coordinates": [
          -0.1359069395277004,
          51.512461774011285
        ],
        "type": "Point"
      }
    },
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "coordinates": [
          -0.1378256466718426,
          51.51303126270315
        ],
        "type": "Point"
      }
    }
  ]
}
```

cdccollins:
> {
>   "type": "FeatureCollection",
>   "features": [
>     {
>       "type": "Feature",
>       "properties": {},
>       "geometry": {
>         "coordinates": [
>           -0.13708768238623747,
>           51.51141463106259
>         ],
>         "type": "Point"
>       }
>     },
>     {
>       "type": "Feature",
>       "properties": {},
>       "geometry": {
>         "coordinates": [
>           -0.1359069395277004,
>           51.512461774011285
>         ],
>         "type": "Point"
>       }
>     },
>     {
>       "type": "Feature",
>       "properties": {},
>       "geometry": {
>         "coordinates": [
>           -0.1378256466718426,
>           51.51303126270315
>         ],
>         "type": "Point"
>       }
>     }
>   ]
> }

Thanks for looking! I think this comes down to my not knowing about geojson stuff... Do you normally pass latitude then longitude, or longitude then latitude? I've assumed that latitude is always first

src/components/my-map/index.ts
```diff
@@ -147,7 +151,32 @@ export class MyMap extends LitElement {
...
+      },
+      {
+        type: "Feature",
+        properties: {},
+        geometry: {
+          coordinates: [-0.1378256466718426, 51.51303126270315],
+          type: "Point",
+        },
+      },
+    ],
```
jessicamcinchak: Oops let's keep default features `[]` here please !

jessicamcinchak (commented):
Aha no worries - you're right it's just an ordering thing - if I switch locally to this then I see markers :white_check_mark: 
```
showNewMarker(
  feature.geometry.coordinates[0],
  feature.geometry.coordinates[1],
);
```

Lat lon :crossed_swords: lon lat is a whole thing in gis mapping land, it's definitely not just you!! https://observablehq.com/@clhenrick/lat-lon-or-lon-lat

src/components/my-map/index.ts
```diff
@@ -604,6 +603,21 @@ export class MyMap extends LitElement {
...
+    };
+
+    // show a marker at a point
+    if (this.showCentreMarker) {
+      showNewMarker(this.markerLongitude, this.markerLatitude);
+    }
+
+    if (this.showGeojsonDataMarkers) {
+      this.geojsonData.features.forEach((feature: GeoJSONFeatureCollection) => {
+        console.log(feature.geometry.coordinates);
```
jessicamcinchak, line 615: sorry two more minor things:
- please remove `console.log()`
- I think `geojsonData` is the type `GeoJSONFeatureCollection` but the singular iterator `feauture` is just `GeoJSONFeature` !

cdccollins: Sorry 🤦‍♀️

jessicamcinchak (approved):
Working now & good to go by me - thanks again :raised_hands: 

Please squash & merge this whenever you're ready & I'll queue up a new release this afternoon and ping you once published to NPM :slightly_smiling_face:

PR #442, opened 2024-04-04
fix: usability improvements

jessicamcinchak:
A number of styling tweaks based on feedback from Planx's most recent accessibility audit: 
- [x] Control button focus color is now GOV.UK yellow for stronger contrast
- [x] Attribution link focus color is now GOV.UK yellow for stronger contrast
- [x] Tab order of control buttons and attribution links is now top to bottom, left to right
- [x] Tab index is omitted from map altogether in cases where `staticMode` is set and attributions are _not_ collapsed

chrome-capture-2024-3-4

Future scope / followup:
- [ ] How to set `aria-label` on canvas element in shadow dom, _not_ container div rendered by Lit
- [ ] Snap points color needs stronger contrast (another basemap altogether?)

### DIFF
--- a/src/components/my-map/index.ts
+++ b/src/components/my-map/index.ts
@@ -364,22 +364,15 @@ export class MyMap extends LitElement {
       node.setAttribute("aria-label", node.getAttribute("title") || ""),
     );
 
-    // Apply tabindexes to OL Controls & Attribution links for accessibility (container div has starting tabindex="1")
-    const controlsLength = olControls?.length;
-    olControls?.forEach((node, i) =>
-      // if `collapseAttributions` is set, ensure attributions control button is last, else follow top-down order
-      node.title === "Attributions"
-        ? (node.tabIndex = 2 + controlsLength)
-        : (node.tabIndex = i + 2),
+    // Re-order overlay elements so that OL Attribution is final element
+    //   making OL Controls first in natural tab order for accessibility
+    const olAttribution = this.renderRoot?.querySelector(
+      ".ol-attribution",
+    ) as Node;
+    const olOverlay = this.renderRoot?.querySelector(
+      ".ol-overlaycontainer-stopevent",
     );
-
-    if (!this.collapseAttributions) {
-      const olAttributionLinks: NodeListOf<HTMLAnchorElement> | undefined =
-        this.renderRoot?.querySelectorAll(".ol-attribution a");
-      olAttributionLinks?.forEach(
-        (node, i) => (node.tabIndex = i + 2 + controlsLength),
-      );
-    }
+    olOverlay?.append(olAttribution);
 
     // define cursors for dragging/panning and moving
     map.on("pointerdrag", () => {
@@ -651,7 +644,7 @@ export class MyMap extends LitElement {
       <div
         id="${this.id}"
         class="map"
-        tabindex="${this.staticMode && !this.collapseAttributions ? -1 : 1}"
+        tabindex="${this.staticMode && !this.collapseAttributions ? -1 : 0}"
       />`;
   }
 
--- a/src/components/my-map/index.ts
+++ b/src/components/my-map/index.ts
@@ -309,22 +309,6 @@ export class MyMap extends LitElement {
     );
     const modify = configureModify(this.drawPointer, this.drawColor);
 
-    // add custom scale line and north arrow controls to the map
-    let scale: ScaleLine;
-    if (this.showNorthArrow) {
-      map.addControl(northArrowControl());
-    }
-
-    if (this.showScale) {
-      scale = scaleControl(this.useScaleBarStyle);
-      map.addControl(scale);
-    }
-
-    if (this.showPrint) {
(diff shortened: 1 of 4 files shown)


jessicamcinchak:
[Map focus order [Usability]](https://trello.com/c/vYhTz9wW/2841-map-focus-order-usability)

jessicamcinchak:
[Map focus [Usability]](https://trello.com/c/ptolMAii/2838-map-focus-usability)

src/components/my-map/index.ts
```diff
@@ -309,22 +309,6 @@ export class MyMap extends LitElement {
...
-    if (this.showScale) {
-      scale = scaleControl(this.useScaleBarStyle);
-      map.addControl(scale);
-    }
-
-    if (this.showPrint) {
-      const printControl = new PrintControl({ map });
-      map.addControl(printControl);
-    }
-
```
jessicamcinchak, line 327: Just adjusting the render order of these so that they come _after_ `handleReset`, consistent with visual display.

ianjon3s:
Looks like a great improvement from a visual pov!

One small suggestion, as the styles are written with SCSS, set a variable (i.e `$color-focus: #ffdd00;`) to reuse the focus colour.

jessicamcinchak:
Thanks @ianjon3s - good spot, now updated! I'm out of practice with stylesheets - always appreciate these tips :slightly_smiling_face:

ianjon3s (approved):
Approving from a non-typescript perspective (@DafyddLlyr may want to add other comments), will be good to see a working example of this to test focus states etc.

src/components/my-map/index.ts
```diff
@@ -357,13 +341,46 @@ export class MyMap extends LitElement {
...
     // Apply aria-labels to OL Controls for accessibility
     const olControls: NodeListOf<HTMLButtonElement> | undefined =
       this.renderRoot?.querySelectorAll(".ol-control button");
     olControls?.forEach((node) =>
       node.setAttribute("aria-label", node.getAttribute("title") || ""),
     );
 
+    // Apply tabindexes to OL Controls & Attribution links for accessibility (container div has starting tabindex="1")
+    const controlsLength = olControls?.length;
+    olControls?.forEach((node, i) =>
```
DafyddLlyr, line 369: I think that the order of these in the DOM is determined by their order in the collection of map controls.

We set up defaults, then based on props, append new ones. This is why the variable ones come after the defaults (e.g. attribution).

Totally fair if you've already explored this one, but it might be worth trying to just manage this via the order they're added to the map controls? Main motivations here would be simplicity (arguable depending on how that works out!), and also the general recommendation to to not use `tabindex` unless totally required - 


> Warning: You are recommended to only use 0 and -1 as tabindex values. Avoid using tabindex values greater than 0 and CSS properties that can change the order of focusable HTML elements ([Ordering flex items](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flexible_box_layout/Ordering_flex_items)). Doing so makes it difficult for people who rely on using keyboard for navigation or assistive technology to navigate and operate page content. Instead, write the document with the elements in a logical sequence.

Source: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

src/components/my-map/index.ts
```diff
@@ -631,7 +648,11 @@ export class MyMap extends LitElement {
   render() {
     return html`<script src="https://cdn.polyfill.io/v2/polyfill.min.js"></script>
       <link rel="stylesheet" href="https://cdn.skypack.dev/ol@^6.6.1/ol.css" />
-      <div id="${this.id}" class="map" tabindex="0" />`;
+      <div
+        id="${this.id}"
+        class="map"
+        tabindex="${this.staticMode ? -1 : 1}"
```
DafyddLlyr:
```suggestion
        tabindex="${this.staticMode ? -1 : 0}"
```

See previous comment referencing MDN docs

DafyddLlyr (commented):
Great to see tests added as always! 

A few suggestions here - possibly something you've already tried.

src/components/my-map/index.ts
jessicamcinchak: This is sorted now by re-ordering, no longer having to rely on positive tab indexes! 

```
    // Re-order overlay elements so that OL Attribution is final element
    //   making OL Controls first in natural tab order for accessibility
    const olAttribution = this.renderRoot?.querySelector(".ol-attribution") as Node;
    const olOverlay = this.renderRoot?.querySelector(".ol-overlaycontainer-stopevent");
    olOverlay?.append(olAttribution);
```

src/components/my-map/index.ts
jessicamcinchak: Also now fixed as result of above and using `-1 : 0` :white_check_mark:

src/components/my-map/index.ts
DafyddLlyr: Great solution - super simple 👍

PR #445, opened 2024-04-08
feat: add optional prop to set `aria-label` on OL canvas element in shadow root

jessicamcinchak:
Adds an optional string prop `ariaLabelOlFixedOverlay` which sets an `aria-label` on the `canvas` element rendered in the shadow root. 

**Per page 98 of recent audit:** 
> The map feature on the page does not have an accessible name. This instance can be found
with the canvas below the ‘about the property’ heading. The map has been written with an
appropriate element, however the element is missing an accessible name, meaning that
screen reader users get relayed the text ‘basemap subject to crown copyright…’, which does
not provide these users with information that is presented visually.
>
> **Solution:**
> As the map contains a focusable element, consider adding an accessible name to the map
feature by adding an aria-label to the canvas.

**References / possible gotchas:** 
- In our first audit, we'd wrongly introduced (and later rolled back) an `ariaLabel` prop on the container div element https://github.com/theopensystemslab/map/pull/86
- There's a stale openlayers issue related to setting this directly on the main canvas (see comment below about _which_ canvas this label is actually being set on): https://github.com/openlayers/openlayers/issues/12485


src/components/my-map/index.ts
```diff
@@ -629,6 +632,10 @@ export class MyMap extends LitElement {
       });
     }
 
+    // Add an aria-label to the overlay canvas for accessibility
+    const olCanvas = this.renderRoot?.querySelector("canvas.ol-fixedoverlay");
```
jessicamcinchak, line 636: There's two `canvas` elements rendered by OL in the shadow root, but I've only been able to successfully select this one - which does differ than the code reference given in our report: 

> **Current code ref(s):**
> #property-information-map > div > div.ol-unselectable.ol-layers > div:nth-child(2) > canvas
<canvas width="1432" height="179" style="position: absolute; left: 0px; transform-
origin: left top; transform: matrix(0.666667, 0, 0, 0.666667, 0, 0);"></canvas>

So, 
- Am I missing something obvious to be able to select the `.ol-layers canvas` element? 
- If that canvas is actually unselectable, are we still happy applying this proposed fix to this canvas and seeing how it re-tests?

jessicamcinchak:
[Interactive map feature [Usability]](https://trello.com/c/KzweGKMG/2837-interactive-map-feature-usability)

src/components/my-map/index.ts
DafyddLlyr: This is a good point to flag and I think I know why you're getting inconsistent results.

My understanding is that OpenLayers doesn't generate either a single canvas, or canvas per-layer, but rather adds additional canvases as needed for memory/rendering. 

I'm trying to find some sort of reference to this online and nothing yet to confirm this in writing!

Looking through OpenLayers examples and inspecting the DOM you can see maps with [multiple layers on a single canvas](https://openlayers.org/en/latest/examples/sea-level.html), and some [spread over multiple canvases](https://openlayers.org/en/latest/examples/layer-opacity.html).

What might be best here is trying to select all `canvas` elements (`querySelectorAll()` I think) and appending `${ariaLabelOlFixedOverlay}-${i}`?

src/components/my-map/index.ts
```diff
@@ -217,6 +217,9 @@ export class MyMap extends LitElement {
     },
   };
 
+  @property({ type: String })
+  ariaLabelOlFixedOverlay = "";
```
DafyddLlyr, line 221: nit: As this is an accessibility related fix, we might be best making this required (and therefore a breaking change)?

This is a little tricky with Lit I think unless we add error handling around it.

Maybe the best option is to making is required in TS types, and throw a `console.warn()` if unset for a few months/versions?

DafyddLlyr (changes requested):
One thing I noticed is that the old `ariaLabel` prop is still mentioned in the Pitsby docs, so it's worth trying to update this at the same time here I think.

image

src/components/my-map/index.ts
jessicamcinchak: Because we also have the new piece of feedback that non-interactive instances of the map (eg PropertyInfo, Review) should _not_ be focusable/in the tab order of the page at all, I was thinking that's a good reason _not_ to require this prop? Eg we don't want to label a non-interactive element (existing visually hidden text should suffice in those scenarios)?

DafyddLlyr: Aha - yep that makes perfect sense, thanks for the context here.

src/components/my-map/index.ts
jessicamcinchak: Dev call resolution here: 
- We don't want to target all & overread (even if we could successfully select all)
- By targeting `fixedoverlay`, we are targetting the outermost canvas which will hopefully still achieve same result as original suggestion
- It's a usability issue, not A-level - so let's keep it timeboxed!

DafyddLlyr (approved):

PR #454, opened 2024-06-27
add role to map

gabimartini:
This PR is relate to the missing role in the div wrapping the map that was found on the process of addressing accessibility concerns on the DPR/Public Index

### DIFF
--- a/src/components/my-map/index.ts
+++ b/src/components/my-map/index.ts
@@ -663,6 +663,7 @@ export class MyMap extends LitElement {
       <div
         id="${this.id}"
         class="map"
+        role="application"
         tabindex="${this.staticMode && !this.collapseAttributions ? -1 : 0}"
       />`;
   }
--- a/src/components/my-map/index.ts
+++ b/src/components/my-map/index.ts
@@ -663,7 +663,9 @@ export class MyMap extends LitElement {
       <div
         id="${this.id}"
         class="map"
-        role="application"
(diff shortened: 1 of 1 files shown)


src/components/my-map/index.ts
```diff
@@ -663,6 +663,7 @@ export class MyMap extends LitElement {
       <div
         id="${this.id}"
         class="map"
+        role="application"
```
jessicamcinchak, line 666: While `role="application"` makes perfect sense for interactive maps, I do wonder if it is applicable when this same component is implemented in "static mode" - eg no interaction, hidden buttons, and excluded from tab index? 

With that scenario in mind, should `role` perhaps be conditional like this? 
```
role={this.staticMode && !this.collapseAttributions ? "presentation" : "application"}
```

Would be curious if you've tested both scenarios or just an interactive one?

jessicamcinchak (approved):
Thanks for this accessibility improvement & contribution! Overall looks great, left one minor question! 

Once merged, I can queue up a new NPM release including this change :rocket:

src/components/my-map/index.ts
gabimartini: Thanks! That's make sense! I made the change and tested with both options, all looking good :)

jessicamcinchak (approved):

src/components/my-map/index.ts
jessicamcinchak: Super, thanks for updating & testing both!

PR #464, opened 2024-07-23
feat: support option to draw & modify regular shapes (box and circle)

jessicamcinchak:

PR #465, opened 2024-07-25
feat: add support for drawing and modifying `Circle` geometry types

jessicamcinchak:
Replaces #464 - let's introduce one new shape at a time ! 

chrome-capture-2024-6-26

**Changes:**
- `drawType` prop now supports `Circle` in addition to original `Polygon` (still default) and `Point`
  - We anticipate using this in Planx to mark the location of trees
- Supports modifying the circle simply by grabbing any edge and dragging - there will _not_ be visible vertice cues like we have with polygon modify
- Refactors styling helper methods and swaps out all `drawType` ternaries to switch statements, which will scale a lot nicer as we support more draw types going forward (for now, the circle is sharing stroke and fill styles with the polygon)

**Followup PRs:**
- Dispatch geojson & area change events when `drawType="Circle"` 
  - "Circle" is not a supported GeoJSON type, so we'll transform it to a polygon first to dispatch - 
  - We should be able to use an OL helper method like [`fromCircle()`](https://openlayers.org/en/latest/apidoc/module-ol_geom_Polygon.html#.fromCircle) here, but getting some type errors (eg our drawn circle is type "SimplifiedCircle" geometry rather than true "Circle" that will be easier to troubleshoot in a separate followup!)

### DIFF
--- a/src/components/my-map/drawing.ts
+++ b/src/components/my-map/drawing.ts
@@ -1,38 +1,41 @@
 import { MultiPoint, MultiPolygon, Polygon } from "ol/geom";
+import { Type } from "ol/geom/Geometry";
 import { Draw, Modify, Snap } from "ol/interaction";
 import { Vector as VectorLayer } from "ol/layer";
 import { Vector as VectorSource } from "ol/source";
 import { Circle, Fill, RegularShape, Stroke, Style } from "ol/style";
 import CircleStyle from "ol/style/Circle";
-import { StyleLike } from "ol/style/Style";
 import { pointsSource } from "./snapping";
 
-export type DrawTypeEnum = "Polygon" | "Point"; // ref https://openlayers.org/en/latest/apidoc/module-ol_geom_Geometry.html#~Type
+export type DrawTypeEnum = Extract<Type, "Polygon" | "Point" | "Circle">;
 export type DrawPointerEnum = "crosshair" | "dot";
 
-// drawPointer styles
-function crosshair(drawColor: string) {
-  return new RegularShape({
-    stroke: new Stroke({
-      color: drawColor,
-      width: 2,
-    }),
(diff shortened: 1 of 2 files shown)


jessicamcinchak:
[Allow applicants to 'work to trees' to see and add more information about protected trees on a map (separate component)](https://trello.com/c/JafdzU94/2855-allow-applicants-to-work-to-trees-to-see-and-add-more-information-about-protected-trees-on-a-map-separate-component)

src/components/my-map/drawing.ts
```diff
@@ -62,60 +64,34 @@ function polygonVertices(drawColor: string) {
...
-function configurePointDrawStyle(pointColor: string) {
-  return new Style({
-    fill: new Fill({ color: pointColor }),
-  });
+  switch (drawType) {
+    case "Point":
+      return new Style({
+        image: new Circle({
+          radius: 9,
+          fill: new Fill({ color: pointColor }),
```
RODO94, line 78: nit: are ``pointColor`` and ``drawColor`` controlling similar things but relate semantically to different parts. Point referring to the color of your _Pointer_ compared with ``drawColor`` which refers to the colour of polygons and shapes which you draw?

Could be cleaner to consolidate them into ``drawColor`` unless they are controlling different things.

RODO94 (approved):
Looks good from what I can see and test, only one wee nit.

src/components/my-map/drawing.ts
jessicamcinchak: Yep totally agree it'd be nice to consolidate everything into `drawColor` ! 

This will technically constitute a breaking change though from the current release - so I vote we handle this in a separate PR where we can adjust behavior and consider marking existing props as `@deprecated` for upcoming release before straight up removing?

PR #466, opened 2024-07-31
feat: add support for drawing many features on the same map

jessicamcinchak:
**Changes:**
- Adds a new `drawMany` prop which has default value `false`
- Works for `drawType` "Polygon" or "Point"
  - Displays labels when `drawMany` is true (label text is auto-index)
- Updates `geojsonChange` event to set `properties`
  - `label` will match visual text on map for any `drawType`
  - `area` replaces previously separate `areaChange` event when `drawType="Polygon"` - see comment
- Also ran "organise imports" and linting on the main `index.ts` - sorry for extra noise!  

**Future scope / follow-up PRs:**
- "Reset" button clears all features as current behavior and resets to empty map, whereas we probably want to consider ability to remove one-by-one (or "last feature" undo?)
  - If removing out of order, should we shift label index or keep static?


jessicamcinchak:
[Allow applicants to 'work to trees' to see and add more information about protected trees on a map (separate component)](https://trello.com/c/JafdzU94/2855-allow-applicants-to-work-to-trees-to-see-and-add-more-information-about-protected-trees-on-a-map-separate-component)

src/components/my-map/index.ts
```diff
@@ -486,23 +487,30 @@ export class MyMap extends LitElement {
...
         if (sketches.length > 0) {
-          const lastSketchGeom = last(sketches)?.getGeometry();
+          if (this.drawType === "Polygon") {
+            sketches.forEach((sketch) => {
+              const sketchGeom = sketch.getGeometry();
+              if (sketchGeom) {
+                sketch.set("area", formatArea(sketchGeom, this.areaUnit));
+              }
+            });
+          }
```
jessicamcinchak, line 503: This is breaking change :warning: rather than dispatching a separate `areaChange` event when drawing polygons, this proposes simply setting `area` within the geojson `properties` and dispatching the single `geojsonChange` event.

A couple benefits:
- Scales much more nicely between draw-one and draw-many
- End consumers only have to subscribe to a single thing and just parse different data from it
- I can't think of any current use cases where you'd _only_ subscribe to an `areaChange` and not care about the whole `geojsonChange` ??

DafyddLlyr: Yeah this is super interesting. I think that this approach will work well and suit our requirements - I can't really think of an issue here. 

Right now we just need the general concept of a change event, and don't really need granular details (e.g. to inform a consumer that the area property of feature 14 have updated). It's enough to say there's been an update to the FeatureCollection as a whole, and it will need to be updated. It's possible that in the future this could be required (seems unlikely and parsable by consumers anyway?), but for now a single "change" event is a lot simpler and meets all requirements 👍

src/components/my-map/drawing.ts
```diff
@@ -64,11 +65,28 @@ function getVertices(drawColor: string) {
...
+    placement: drawType === "Point" ? "line" : "point", // "point" placement is center point of polygon
+    fill: new Fill({
+      color: "#000",
+    }),
+    stroke: new Stroke({
+      color: "#fff",
+      width: 3,
+    }),
+  });
+}
```
DafyddLlyr, line 81: Perfect!

DafyddLlyr (approved):
Nice change with a very small diff - love to see it 🙌

PR #473, opened 2024-08-02
break: condense and deprecate redundant styling-related props

jessicamcinchak:
Our styling-related props haven't always been so consistent, so let's tidy them up with a v1 release in mind & flag some as deprecated. 

Here's a quick overview of exisitng & changing style-related props based on various modes/functionalities:
- **Drawing**
  - `drawColor` (string) - fill automatically applied at 20% (used to be 10%)
  - ~`drawPointColor` (string)~ now deprecated - it's only possible to set a single `drawType` at a time anyways, so `drawColor` is sufficient
  - ~`drawFillColor` (string)~ now deprecated
  - `drawPointer` (enum)
- **Showing a center map marker** (based on lat,lng props _not_ geojson)
  - `markerColor` (string)
  - `markerImage` (enum)
- **Displaying GeoJSON data**
  - `geojsonColor` (string)
  - `geojsonFill` (boolean) - 20% if true
- **Showing OS Features at point**
  - `featureColor` (string)
  - `featureFill` (boolean) - 20% if true
  - ~`featureBorderNone` (string)~ now deprecated
  
Any other votes for deprecation or refactoring?

### DIFF
--- a/src/components/my-map/drawing.ts
+++ b/src/components/my-map/drawing.ts
@@ -1,3 +1,4 @@
+import { FeatureLike } from "ol/Feature";
 import { MultiPoint, MultiPolygon, Polygon } from "ol/geom";
 import { Type } from "ol/geom/Geometry";
 import { Draw, Modify, Snap } from "ol/interaction";
@@ -6,19 +7,10 @@ import { Vector as VectorSource } from "ol/source";
 import { Circle, Fill, RegularShape, Stroke, Style, Text } from "ol/style";
 import CircleStyle from "ol/style/Circle";
 import { pointsSource } from "./snapping";
-import { FeatureLike } from "ol/Feature";
+import { hexToRgba } from "./utils";
 
 export type DrawTypeEnum = Extract<Type, "Polygon" | "Point" | "Circle">;
 export type DrawPointerEnum = "crosshair" | "dot";
-
-function getFillFromColor(drawColor: string) {
-  if (drawColor.startsWith("#")) {
-    // 10% opacity
-    return "#1A" + drawColor.substring(1);
-  } else {
-    return drawColor;
-  }
-}
 
 function configureDrawPointerImage(
   drawPointer: DrawPointerEnum,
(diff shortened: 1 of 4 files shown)


src/components/my-map/drawing.ts
```diff
@@ -10,6 +10,15 @@ import { pointsSource } from "./snapping";
 export type DrawTypeEnum = Extract<Type, "Polygon" | "Point" | "Circle">;
 export type DrawPointerEnum = "crosshair" | "dot";
 
+function getFillFromColor(drawColor: string) {
+  if (drawColor.startsWith("#")) {
+    // 10% opacity
+    return "#1A" + drawColor.substring(1);
```
jessicamcinchak, line 16: todo: open layers doesn't seem to recognise this format - need to try hex to rgba instead?

src/components/my-map/index.ts
```diff
@@ -88,6 +88,9 @@ export class MyMap extends LitElement {
   @property({ type: String })
   drawType: DrawTypeEnum = "Polygon";
 
+  /**
+   * @deprecated - please set `drawColor` regardless of `drawType`
+   */
```
jessicamcinchak, line 93: My thinking is we'll do a NPM `prerelease` _with_ the `@deprecated` annotations even though the types are no longer being read in anywhere, then fully remove them ahead of the stable release - sound like a decent plan?

DafyddLlyr: Perfect!

src/components/my-map/index.ts
```diff
@@ -418,9 +421,8 @@ export class MyMap extends LitElement {
     const geojsonLayer = new VectorLayer({
       source: geojsonSource,
       style: function (this: MyMap, feature: any) {
-        // Retrieve color from feature properties
-        let featureColor = feature.get("color") || this.geojsonColor; // Use the geojsonColor if no color property exists
-
+        // Read color from geojson feature `properties` if set or fallback to prop
+        let featureColor = feature.get("color") || this.geojsonColor;
```
DafyddLlyr, line 425: nit: It looks like this could actually be a `const`?

Another "pre v1" task might be to set up linting here to catch a few of these type of things?

DafyddLlyr (approved):
Super helpful and clear PR description thanks 🙌

src/components/my-map/index.ts
jessicamcinchak: Yep great shout - will handle in a followup PR :+1:

PR #474, opened 2024-08-06
build(deps): upgrade Open Layers

jessicamcinchak:
Ran `pnpm upgrade ol ol-ext ol-mapbox-style @types/ol-ext`

A number of type errors to sort out! Especially still need to debug / adjust a breaking change to our Snap interaction's `pointsSource` per changelog here: https://github.com/openlayers/openlayers/releases/tag/v9.2.0

See current type errors:
```
Type 'VectorSource<never>' is not assignable to type 'VectorSource<Feature<Geometry>>'.
  Types of property 'setLoader' are incompatible.
```

Then might try to get us all the way up to v10.x


DafyddLlyr:
Replaced by https://github.com/theopensystemslab/map/pull/535 and

PR #479, opened 2024-08-24
fix: update `featuresGeojsonChange` event to match new `geojsonChange` behavior

jessicamcinchak:
Follows on from #466 

**Changes:**
- Similar to `drawMode` - the `clickFeatures` functionality dispatches geojson & area events which should stay consistent with recent changes we've made to event dispatching - so we're deprecating `featuresAreaChange` event and instead setting `"area"` directly on the `featuresGeojsonChange` event data "properties"
- Deprecates `areaUnit` prop and simply sets _both_ `"area.squareMetres"` and `"area.hectares"` anywhere we calculate the area (these field names & new number type are more consistent with ODP Schema too)

### DIFF
--- a/src/components/my-map/index.ts
+++ b/src/components/my-map/index.ts
@@ -54,7 +54,7 @@ import styles from "./styles.scss?inline";
 import {
   AreaUnitEnum,
   fitToData,
-  formatArea,
+  calculateArea,
   hexToRgba,
   makeGeoJSON,
 } from "./utils";
@@ -234,6 +234,9 @@ export class MyMap extends LitElement {
   @property({ type: Boolean })
   staticMode = false;
 
+  /**
+   * @deprecated - both `area.squareMetres` & `area.hectares` are carculated by defualt now in applicable `geojsonChange` events
+   */
   @property({ type: String })
   areaUnit: AreaUnitEnum = "m2";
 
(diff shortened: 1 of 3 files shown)


src/components/my-map/index.ts
```diff
@@ -234,6 +234,9 @@ export class MyMap extends LitElement {
   @property({ type: Boolean })
   staticMode = false;
 
+  /**
+   * @deprecated - both `area.squareMetres` & `area.hectares` are carculated by defualt now in applicable `geojsonChange` events
```
DafyddLlyr, line 238:
```suggestion
   * @deprecated - both `area.squareMetres` & `area.hectares` are calculated by default now in applicable `geojsonChange` events
```

src/components/my-map/index.ts
```diff
@@ -555,10 +549,15 @@ export class MyMap extends LitElement {
...
+            // Calculate the "area" and set on geojson `properties`
             sketches.forEach((sketch) => {
               const sketchGeom = sketch.getGeometry();
               if (sketchGeom) {
-                sketch.set("area", formatArea(sketchGeom, this.areaUnit));
+                sketch.set(
+                  "area.squareMetres",
+                  calculateArea(sketchGeom, "m2"),
+                );
+                sketch.set("area.hectares", calculateArea(sketchGeom, "ha"));
```
DafyddLlyr, line 560: fyi: There's a `setProperties()` that allows multiple values to be set at once. No better or worse, just sharing!

src/components/my-map/utils.ts
```diff
@@ -13,25 +13,26 @@ import { ProjectionEnum } from "./projections";
...
-  let output;
-  if (unit === "m2") {
-    output = squareMetres + " m²";
-  } else if (unit === "ha") {
-    output = hectares + " ha";
+  switch (unit) {
+    case "m2":
+      return squareMetres;
+    case "ha":
+      return hectares;
```
DafyddLlyr, line 34: Nice!

DafyddLlyr (approved):

src/components/my-map/index.ts
jessicamcinchak: thank you :white_check_mark: :sweat_smile:

PR #483, opened 2024-08-29
fix: increase `drawMany` label font size and default drawing point size

jessicamcinchak:
After screenshots:
Screenshot from 2024-08-29 12-37-21
Screenshot from 2024-08-29 12-35-56

### DIFF
--- a/src/components/my-map/drawing.ts
+++ b/src/components/my-map/drawing.ts
@@ -69,14 +69,14 @@ function getVertices(drawColor: string) {
 function styleFeatureLabels(drawType: DrawTypeEnum, feature: FeatureLike) {
   return new Text({
     text: feature.get("label"),
-    font: "80px inherit",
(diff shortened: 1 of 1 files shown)


src/components/my-map/drawing.ts
```diff
@@ -69,14 +69,14 @@ function getVertices(drawColor: string) {
 function styleFeatureLabels(drawType: DrawTypeEnum, feature: FeatureLike) {
   return new Text({
     text: feature.get("label"),
-    font: "80px inherit",
```
jessicamcinchak, line 72: Realised "inherit" actually isn't supported here, so was just falling back to the OL default - hence `20px` change is actually _bigger_ than original `80px` defined here :see_no_evil:

DafyddLlyr (approved):
Nice!

PR #491, opened 2024-09-02
fix: `drawGeojsonData` should accept "Feature" and "FeatureCollection" data since introducing `drawMany`

jessicamcinchak:


src/components/my-map/index.ts
```diff
@@ -460,23 +460,23 @@ export class MyMap extends LitElement {
...
       let features = new GeoJSON().readFeatures(this.geojsonData, {
         featureProjection: "EPSG:3857",
       });
       geojsonSource.addFeatures(features);
-      geojsonSource.setAttributions(this.geojsonDataCopyright);
     } else if (this.geojsonData.type === "Feature") {
       let feature = new GeoJSON().readFeature(this.geojsonData, {
         featureProjection: "EPSG:3857",
       });
       geojsonSource.addFeature(feature);
```
RODO94, line 475: Took me a second to realise the differences between these parts. Is the need for two different functions relate to one accessing the `Features` array of a `FeatureCollection` and the other just taking the `Feature` as an object, or is there another reason for having two different functions / if clauses?

RODO94 (approved):
Looks good to me to switch from pulling in a feature to using a FeatureCollection like in geoJsonData prop. 

Only had one question relating to adding a Feature or Features, mostly curious about the structuring of the code.

src/components/my-map/index.ts
jessicamcinchak: Yes exactly OpenLayers provides different methods depending if GeoJSON is single Feature or a FeatureCollection, and we want to keep our public-facing prop flexible to accepting eithe, therefore we do this conditional check !

PR #498, opened 2024-09-06
style: Reset icon direction reverse

ianjon3s:
## What does this PR do?

Updates direction of reset icon (`↺`) to be consistent with global PlanX reset button.


jessicamcinchak (approved):
Perfect !

PR #507, opened 2024-10-17
style: Improve legibility of dot labels

ianjon3s:
## What does this PR do?

Improves visibility of point labels by removing character stroke and keeping background/text colour constant (not dependent on custom colour selection).

**Preview (before vs after):**
image

### DIFF
--- a/src/components/my-map/drawing.ts
+++ b/src/components/my-map/drawing.ts
@@ -89,11 +89,11 @@ function configureDrawingLayerStyle(
     case "Point":
       return new Style({
         image: new Circle({
-          radius: 12,
+          radius: hideDrawLabels ? 10 : 12,
           fill: new Fill({ color: "#fff" }),
           stroke: new Stroke({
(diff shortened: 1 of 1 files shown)


jessicamcinchak (changes requested):
Hi ! I think this looks great for `drawMany` cases when labels are visible - but can we please confirm this is still the intended style for:
- drawing a single point
- `drawMany` _with_ `hideDrawLabels` as recently introduced in #508

I think I'd still expect to still see a solid colored point in these cases rather than the stroke only?

Screenshot from 2024-10-21 10-26-25

ianjon3s:
Good point @jessicamcinchak , however we still want to show two contrasting colours to safeguard against any contrast issues with the background map (green on green < green+white on green). I've updated the stroke to be thicker in the absence of labels:

image

jessicamcinchak (approved):
Perfect thanks for update!

PR #509, opened 2024-10-23
fix: consistent point styles

jessicamcinchak:
Quick follow-on to #507 

- Ensures drawing a _single_ point uses same style as `drawMany` with hidden labels
- Bumps radius of center marker & geojson data points to match drawing points


RODO94 (approved):
The refinements look good to me, thanks for picking these up

PR #538, opened 2025-03-20
chore: CVE-2025-29907

DafyddLlyr:
Fixes -
- https://github.com/theopensystemslab/map/security/dependabot/64 
- https://github.com/theopensystemslab/map/security/dependabot/63


freemvmt (approved):
Nice

PR #570, opened 2025-08-08
fix: output 4326 for lookup against planning.data.gov.uk API

jmcbroom:
- output 4326 coordinates

belatedly realizing we need to pass lat/lng to the next API here, so adding that to the output;

the BNG coordinates are still present in the output if those need to be used anywhere.


jessicamcinchak (approved):
Good catch! Once you merge this, happy to do another quick package release here (npm publishing permissions are limited, sorry!)

PR #617, opened 2026-04-16
docs: Migrate from Pitsby to Storybook

DafyddLlyr:
## What does this PR do?
 - Removes Pitsby as a docs library - this is no longer supported, and has a number of security vulnerabilities
 - Replaces with Storybook - this matches what we use in PlanX so it seemed like an easy choice

Demo - https://deploy-preview-617--oslmap.netlify.app/

This was my first time fully handing over a task to Claude Code and it wasn't too bad...? It got about 90% of the way there in one shot, then just needed some help across the line. Seems a good tool for tasks like this which we would have struggled to prioritise and dedicate time towards. Happy to talk this through at a dev call sometime if of interest.


jessicamcinchak (approved):
Great ! This indeed a nice bit of debt to have cleaned up ! 

Storybook's "grouped props"  is especially nice here for readability / organisation, I haven't seen this before 🌟 https://deploy-preview-617--oslmap.netlify.app/?path=/docs/components-mymap--docs

PR #618, opened 2026-04-27
feat: optionally show a search bar when drawing on an OS basemap to position the map view

jessicamcinchak:
**Changes / how this works**
- Add a prop `showOSSearch` which will render the geocode autocomplete search component above the map
- The searched address is used to _position_ the map only, it does not inherently place the draw point (because may be using `drawMany`, draw `polygon` etc)
- The geocode autocomplete will inherently return _all_ OS addresses throughout the UK, but an error will be thrown if the map uses a small clip extent (eg Lambeth's boundary) and you search an address outside of it (eg in Buckinghamshire)

**Next steps**
- Merge this and publish a new alpha/@next release of this package https://github.com/theopensystemslab/map/blob/main/CHANGELOG.md
- Bring into planx-new feature branch and hookup + test on FindProperty's "Propose a new address" route

**Prior art**
Originally started here https://github.com/theopensystemslab/planx-new/pull/6455, but ultimately couldn't hack two-way data flow and Lit's `update` life cycle so opted for this "composite component" approach directly within `render` and controlled via map props instead :heavy_plus_sign:

### DIFF
--- a/src/components/my-map/index.ts
+++ b/src/components/my-map/index.ts
@@ -72,6 +72,9 @@ export class MyMap extends LitElement {
   @property({ type: String })
   id = "map";
 
+  @property({ type: Boolean })
+  showOSSearch = false;
+
   @property({ type: String })
   dataTestId = "map-test-id";
 
@@ -288,7 +291,7 @@ export class MyMap extends LitElement {
   }
 
   // runs after the initial render
-  firstUpdated() {
+  async firstUpdated() {
     const target = this.renderRoot.querySelector(`#${this.id}`) as HTMLElement;
 
     const isUsingOS = Boolean(this.osApiKey || this.osProxyEndpoint);
@@ -785,6 +788,41 @@ export class MyMap extends LitElement {
       });
     }
 
+    // Duplicated logic in render()
+    const showSearch =
+      this.showOSSearch &&
+      this.drawMode &&
+      ["OSVectorTile", "OSRaster"].includes(this.basemap) &&
+      (Boolean(this.osApiKey) || Boolean(this.osProxyEndpoint));
+
+    if (showSearch) {
+      const search = this.renderRoot?.querySelector("geocode-autocomplete");
+      if (search) {
+        // Give the browser a chance to paint
+        //  Ref https://lit.dev/docs/v1/components/events/#add-event-listeners-after-first-paint
+        await new Promise((r) => setTimeout(r, 0));
+
+        search.addEventListener(
+          "addressSelection",
+          ({ detail: address }: any) => {
+            console.debug("searched", { detail: address });
+            const searchedAddress = address?.address?.LPI;
+            const newCenterCoordinate = transform(
+              [searchedAddress.LNG, searchedAddress.LAT],
+              "EPSG:4326", // LPI output srs
+              "EPSG:3857",
+            );
+
+            // TODO handle validation here before recentering (eg is searched point within clip extent)
+            //   error wrapper on geocode autocomplete?? or popup/toast on map?
+
+            map.getView().setCenter(newCenterCoordinate);
+            map.getView().setZoom(20);
+          },
+        );
+      }
+    }
+
     // Add an aria-label to the overlay canvas for accessibility
     const olCanvas = this.renderRoot?.querySelector("canvas.ol-fixedoverlay");
     olCanvas?.setAttribute("aria-label", this.ariaLabelOlFixedOverlay);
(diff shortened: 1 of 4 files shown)


src/components/my-map/index.ts
```diff
@@ -269,12 +273,7 @@ export class MyMap extends LitElement {
   collapseAttributions = false;
 
   @property({ type: Object })
-  clipGeojsonData = {
-    type: "Feature",
-    geometry: {
-      coordinates: [],
-    },
-  };
+  clipGeojsonData: GeoJSONFeature | undefined = undefined;
```
jessicamcinchak, line 276: Lit `Object` types have always been a bit weird with default value handling, but this feels a lot more semantically meaningful now and _should be_ functioning exactly the same (eg snap feature which has good unit test coverage also relies on clip extent).

src/components/my-map/index.ts
```diff
@@ -797,21 +840,75 @@ export class MyMap extends LitElement {
...
-          ? "presentation"
-          : "application"}"
-        tabindex="${this.staticMode && !this.collapseAttributions ? -1 : 0}"
-        data-testid="${this.dataTestId}"
-      />`;
+    this._showSearch =
+      this.showOSSearch &&
+      this.drawMode &&
+      ["OSVectorTile", "OSRaster"].includes(this.basemap) &&
+      (Boolean(this.osApiKey) || Boolean(this.osProxyEndpoint));
```
jessicamcinchak, line 858: "Showing search" isn't as simple as setting only the `showOSSearch` prop, we also want to confirm that these other conditions are met for the best UI/UX.

This should avoid a few weird edge cases which could otherwise be created, for example: 
- Showing search on a _static_ map, not in draw mode
- Showing OS-based search on a draw mode map using Mapbox Satellite basemap

src/components/my-map/styles.scss
```diff
@@ -1,3 +1,5 @@
+@import "govuk-frontend/dist/govuk/index";
```
jessicamcinchak, line 1: Now required for styling `govuk` error classes.

src/components/my-map/index.ts
```diff
@@ -797,21 +840,75 @@ export class MyMap extends LitElement {
...
   }
 
+  _showSearchError() {
+    const errorEl: HTMLElement | null | undefined =
+      this.shadowRoot?.querySelector(`#geocode-autocomplete-error`);
+
+    // display "none" ensures always present in DOM, which means role="status" will work for screenreaders
+    if (errorEl) errorEl.style.display = "none";
+    if (errorEl && this._searchError) errorEl.style.display = "";
+  }
```
jessicamcinchak, line 850: This follows the pattern established by `postcode-search` and `address-autocomplete`, so should hopefully be a11y compliant come our next audit 🤞

src/components/my-map/index.ts
DafyddLlyr: Makes sense to me...! 

We have a few similar guards here where props need to be grouped. If this was a pure React component it would be a good case for a discriminated union of props to enforce this, but I'm not too sure how we'd...
 - Set that up in Lit
 - Enforce this in a consumer application (exporting prop types somewhere?)

One to have a look at (timeboxed) in future maybe.

src/components/my-map/my-map.stories.ts
DafyddLlyr, line 1: Great to see this included 👍

src/components/my-map/my-map.stories.ts
```diff
@@ -506,6 +516,24 @@ export const DrawModeGeoJSONOutput: Story = {
...
+  render: () => `
+    <my-map
+      id="draw-mode-with-search"
+      zoom="20"
+      maxZoom="23"
+      drawMode
+      drawPointer="dot"
+      showOSSearch
+      basemap="OSVectorTile"
+      osVectorTilesApiKey="${osApiKey}"
```
DafyddLlyr, line 533: I actually can't get this to render on the Netlify env, or locally, when I pass in either a key or proxy endpoint.

This might be a problem with how I originally set this up though, the same applied to other stories 🤔 

Were you able to get this to work locally?

DafyddLlyr (approved):

src/components/my-map/my-map.stories.ts
jessicamcinchak: Was also not able to get Storybook working locally, even with API key as plain string passed in, but not uniuqe to this new story for me! Happy to come back to this as a follow-up, meant to similarly point out here!  

Local dev & index.html examples were completely fine/working as usual, so not anticipating issues with FindProperty integration. But will be an `alpha` release regardless, justt in case!

PR #626, opened 2026-05-12
chore: Update deps

DafyddLlyr:
Fixes - 
 - https://github.com/theopensystemslab/map/security/dependabot/149
 - https://github.com/theopensystemslab/map/security/dependabot/150


jessicamcinchak (approved):
Thanks for being on top of these, Daf! Merging so included in the next release :checkered_flag:

PR #627, opened 2026-05-14
fix: correct height and error styles for map with `showOSSearch` enabled

jessicamcinchak:
Fixes two noticable issues when using in planx:
- [x] Autocomplete search error message
  - Missing govuk-style left border when visible
  - Not clearing on address selection _change_ 
- [x] Overall height when search + map are grouped 
  - `map` class requires height to render (currently 100% with + without search); this means search + map are > 100% and overlapping other content like planx's `MapFooter`
  - To test / recreate: `pnpm dev` from this branch, see index.html elements overlapping here when map has search visible 
image

### DIFF
--- a/index.html
+++ b/index.html
@@ -55,7 +55,8 @@
           drawMode
           drawMany
           drawType="Point"
-          basemap="MapboxSatellite"
+          basemap="OSVectorTile"
+          showOSSearch
           showCentreMarker
           osCopyright="© Crown copyright and database rights 2024 OS AC0000812160"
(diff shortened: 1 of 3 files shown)


DafyddLlyr:
@jessicamcinchak I believe that [fa09311](https://github.com/theopensystemslab/map/pull/627/commits/fa093110d9e370466c9d4faeb287da3392776180) resolves the issue here..!

image

joecarver (approved):
both sets of fixes seem to be working for me!

PR #637, opened 2026-06-25
feat: Improve styling options for Geocode component

DafyddLlyr:
This PR allows us to style the `<my-map showOSSearch />` component to allow the following - 
 - custom max-width
 - custom font (overwriting default GDS Transport")

It also removes bottom margin and padding which was growing the error wrapper larger than the input.
 
 
image

### DIFF
--- a/src/components/geocode-autocomplete/styles.scss
+++ b/src/components/geocode-autocomplete/styles.scss
@@ -19,6 +19,7 @@
     font-family: $font-family;
     font-size: $label-font-size;
     margin-bottom: $label-margin-bottom;
+    max-width: $input-max-width;
   }
 
   .autocomplete__input {
@@ -51,5 +52,6 @@
 
   .autocomplete__menu {
     max-height: var(--autocomplete__menu__max-height, 342px);
+    max-width: calc($input-max-width - 4px);
   }
 }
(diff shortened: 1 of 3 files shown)


src/components/geocode-autocomplete/styles.scss
```diff
@@ -51,5 +52,6 @@
 
   .autocomplete__menu {
     max-height: var(--autocomplete__menu__max-height, 342px);
+    max-width: calc($input-max-width - 4px);
```
DafyddLlyr, line 55: The 4px here ensures the menu is within the input (accounting for border and focus).

ianjon3s (approved):
Tested locally:

- ✅ Error text Inter font
- ✅ Error text weight semi-bold
- ✅ Error text margin removed
- ✅ Expanding select positioned "within" parent input

One tiny remaining issue (which you mentioned earlier) is the slight difference in red shade from our error text, but it's 99.5% there, which is more than close enough.

DafyddLlyr:
> _One tiny remaining issue (which you mentioned earlier) is the slight difference in red shade from our error text, but it's 99.5% there, which is more than close enough._

This can be overwritten via the CSS var `--govuk-error-colour` (see 🩷 hotpink 🩷  example above) - within PlanX we can just pass in `--govuk-error-colour: theme.palette.error.main`

PR #646, opened 2026-07-02
refactor: Update map bounds error text copy

ianjon3s:
## What does this PR do?

- Updates out-of boundary error on maps to read: 

> This address falls outside the map's coverage area. Please try another


DafyddLlyr (approved):
Hey there - totally missed this follow up sorry. Once merged I'll sort out a publish here and we can bring this into `planx-new`

PR #653, opened 2026-08-24
fix(a11y): Solve bleedthrough issue by use of `tStatusNoResults` property

DafyddLlyr:
## What's the problem?

> **Bleed-through**
> There is content which is visually hidden but not hidden from assistive technology.
>
> Upon landing on the ‘Select an address’ field, a status message is exposed to screen reading
software which announces ‘no search results’. This is incorrect as there are results within
the drop-down.

Source: DAC Accessibility Report 2026, p. 22

I can recreate this - when using the Mac OS screen reader I get "No search results" read out when there are results in the list (see ~28s below) - 

https://github.com/user-attachments/assets/832dab9f-8cee-457c-94e6-31b1b0d7905e

## What's the cause?
As identified by @ianjon3s this is related to the underlying `accessible-autocomplete` library we're building on here. (alphagov/accessible-autocomplete#395).

## What's the solution?
Use the prop `tStatusNoResults` to check the actual state of the data rather than relying on the internal (stale) option count ([docs](https://github.com/alphagov/accessible-autocomplete#tstatusnoresults-default---no-search-results)).

Please see video below which now reads out the number of results directly (~14s) - 

https://github.com/user-attachments/assets/f881abe4-9f1d-452a-a1f9-bb4d81840899

## Testing
I've added a Storybook which mimic the PlanX `FindProperty` setup which allows to start with an empty list, then populate it, which is needed to trigger this edge case.

Please test the issue is not present here - https://deploy-preview-653--oslmap.netlify.app/?path=/story/components-addressautocomplete--find-property

Running `pnpm dev` locally you can uncomment out the `tStatusNoResults` changes to run the story and see the issue in place.


DafyddLlyr:
Trello card - https://trello.com/c/oStiYqOr/3716-a-bleedthrough-01

jessicamcinchak (approved):
Nice & simple! Shout if you want a hand with a new (pre)release, but I think you have access to NPM too?

